PR #84, opened 2026-04-10
fix: revert snake_case frontend types — Go backend sends PascalCase

cedbossneo:
## Summary
- **Reverts #83** — the Go backend has a `snakeToPascalJSON` transform in `providers/transform.go` that converts all rosbridge snake_case JSON to PascalCase before sending to the frontend WebSocket
- The frontend types were already correct with PascalCase field names
- Deployment was broken because all property accesses returned `undefined`

## Root cause
`gui/pkg/providers/transform.go` contains `snakeToPascalJSON()` which converts every map key from snake_case to PascalCase before the data reaches the frontend. The snake_case migration assumed raw rosbridge JSON reached the frontend, but it doesn't.

## Test plan
- [ ] Merge and deploy immediately to restore functionality
- [ ] Verify all WebSocket subscriptions show data (status, GPS, emergency, power, IMU)


cedbossneo:
Closing in favor of removing snakeToPascalJSON transform from the Go backend instead

Issue #40, opened 2026-04-07
feat: headland passes — mow perimeter before interior fill


cedbossneo:
## Description
Mow the perimeter outline (headland) before filling the interior with boustrophedon swaths.

The B-RV planner (`mowgli_brv_planner`) already generates headland outline waypoints in `brv_algorithm.cpp`, but they are only used for visualization — the outline path is not executed by the robot before the fill.

## Requirements
- Execute 2 headland passes along the boundary polygon (inset by tool_width per pass)
- Headland passes should run with blade enabled
- After headland, transition to interior fill (existing boustrophedon sweep)
- Configurable number of passes (`headland_passes` param, default 2)

## Current State
- `offset_polygon_inward()` in `geometry_utils.cpp` generates the headland rings
- `brv_algorithm.cpp` stores them in `result.outline_path` but they are not part of `result.full_path`

## Files
- `ros2/src/mowgli_brv_planner/src/brv_algorithm.cpp`
- `ros2/src/mowgli_behavior/src/coverage_nodes.cpp`

cedbossneo:
Implemented in dev: headland passes via F2C `generateHeadlandSwaths()` (`ros2/src/mowgli_coverage/src/coverage_server.cpp:327`) — concentric perimeter rings prepended to inner-field swaths. See CLAUDE.md § Coverage.

PR #184, opened 2026-05-05
feat(mowing): wire mowing_enabled and angle params into ROS2 nodes

### LINKED ISSUE #142: feat(mowing): wire outline and angle parameters to ROS2
## Problem

The mowing pattern parameters are configurable in the GUI Settings page (Mowing section) and persisted in `mowgli_robot.yaml`, but **no ROS2 node declares or reads them**. Editing them on the GUI looks like it should change the mowing behavior — and silently doesn't.

## Affected parameters

- `mowing_enabled`
- `outline_passes`
- `outline_overlap`
- `outline_offset` / `outline_offset_m`
- `mow_angle_offset_deg`
- `mow_angle_increment_deg`

## Where they should be wired

`map_server_node` (cell-based strip planner) is where the coverage path is generated. The angle/offset/overlap params should drive the strip-pre-plan logic; the outline params should drive the perimeter-pass that runs before the strip phase. `mowing_enabled` likely belongs in `behavior_tree_node` as a master gate before `RecordArea` / `MowArea` ticks emit blade-on commands.

## Acceptance

- [ ] Each param is declared via `declare_parameter<>()` on a real node
- [ ] The launch wiring (`navigation.launch.py` or equivalent) injects the value from `mowgli_robot.yaml` so the GUI Settings → Save flow takes effect after restart
- [ ] Tuning the value on the GUI changes the planned coverage path (verifiable on `/map_server_node/coverage_cells`)

## Discovered via

`docker/config/mowgli/mowgli_robot.yaml` orphan-parameter audit, 2026-05-02.

cedbossneo:
## Summary

Closes #142.

The mowing pattern parameters in \`mowgli_robot.yaml\` (\`mowing_enabled\`, \`mow_angle_offset_deg\`, \`mow_angle_increment_deg\`) were exposed in the GUI Mowing settings but **no ROS2 node read them** — editing them looked like it should change behavior and silently didn't.

### Behavior tree — master mowing toggle
- New \`IsMowingEnabled\` condition node reading from the BT blackboard.
- \`behavior_tree_node\` declares \`mowing_enabled\` (default \`true\`) and pushes it onto the blackboard alongside the existing battery thresholds.
- \`main_tree.xml\`: \`MowingSequence\` gates COMMAND_START on \`IsMowingEnabled\`, so flipping the GUI's *Mowing enabled* switch off makes Start a no-op (the parent Fallback drops to \`IdleSequence\`). Recording, manual mow, critical-battery dock, and Home flows are intentionally **not** gated by this — operators may still need to drive or recover the robot while mowing is globally disabled.

### Map server — strip-pattern angle tuning
- \`map_server_node\` declares \`mow_angle_offset_deg\` and \`mow_angle_increment_deg\` (both default 0).
- \`coverage_planner::ensure_strip_layout\` applies them after the base angle (MBR auto or manual override):
  \`final_angle = base + offset + increment * area_index\`
  Constant offset rotates every area uniformly; increment lets multi-area gardens stagger swath direction.

### Launch wiring
- \`full_system.launch.py\` reads the three keys from \`robot_params\` and injects them as ROS parameters on the right node.

### Defaults
- \`mowgli_robot.yaml\` \`mow_angle_offset_deg: -1\` → \`0\`. The previous \`-1\` was a stale comment about \"auto-detect\" — auto detection is the NaN sentinel on \`mow_angle_deg\`, not the offset.

### Outline params (deliberately not wired)
\`outline_passes\`, \`outline_overlap\`, \`outline_offset\` are listed in #142 too, but the cell-based strip planner has no perimeter-pass implementation yet (that's #40 — headland passes). Wiring them would still be silent. Instead, added an info \`Alert\` banner in the Mowing settings *Perimeter (Outline)* card pointing to #40 so operators know the values they save are retained for when that issue lands.

## Component

- [x] ROS2 (\`ros2/\`)
- [x] GUI (\`gui/\`)

## Testing

- [x] \`yarn tsc --noEmit\` passes
- [x] \`yarn build\` clean
- [x] BT XML \`xmllint --noout\` passes
- [x] Launch file \`python3 -m py_compile\` passes
- [ ] \`colcon build --packages-select mowgli_behavior mowgli_map mowgli_bringup\` — needs a running devcontainer (none up locally right now); CI / on-robot build will validate. Local clang-only diagnostics are noise (no ROS includes on host).
- [ ] On hardware: toggle \`mowing_enabled\` off → press Start → robot stays idle. Set \`mow_angle_offset_deg=15\` → \`/map_server_node/coverage_cells\` shows 15° rotated swaths.

## Notes

The change to \`mowing_enabled: true\` default keeps the existing behaviour for users with no override.


cedbossneo:
Closing as stale — this branch is from May and now conflicts across 100+ files after two months of `dev` churn, making a rebase impractical. If the feature is still wanted, it's cleaner to re-cut it against current `dev`. (Mowing param wiring is still tracked by #142.)

PR #185, opened 2026-05-05
fix(ros2): drop dead mowgli_nav2_plugins COPY lines from Dockerfile

cedbossneo:
## Summary

Unblocks every ROS2 — Docker run on \`feat/opennav-coverage\`.

The \`mowgli_nav2_plugins\` package (FTCController) was deleted in commit \`da71b13 feat(coverage): migrate to opennav_coverage + nav2_mppi_controller\`, but \`ros2/Dockerfile\` still tried to COPY \`src/mowgli_nav2_plugins/{package.xml, CMakeLists.txt}\` as part of the rosdep-cache layer. With those files gone, every Docker build fails at the deps stage:

\`\`\`
ERROR: failed to build: failed to solve: failed to compute cache key:
failed to calculate checksum of ref ...: \"/src/mowgli_nav2_plugins/CMakeLists.txt\": not found
\`\`\`

Confirmed across runs 25395317876, 25397188076, 25397440790, 25397667386, 25397829852, 25398274876, 25398773938, 25400661889, 25402197143 — all failing at the same step.

## Fix

Drop the two COPY lines (lines 160–161 of \`ros2/Dockerfile\`) and leave a comment pointing at \`da71b13\` so the removal isn't accidentally re-introduced. Also nudge the example in \`ros2/scripts/build.sh\` to use a package that still exists.

## Component

- [x] ROS2 (\`ros2/\`) — Dockerfile only

## Testing

- [x] \`grep mowgli_nav2_plugins ros2/Dockerfile\` only finds the comment now.
- [ ] CI: this PR triggering ROS2 — Docker should now reach the build stage cleanly.

## Notes

Stale README references to FTCController / \`mowgli_nav2_plugins\` remain (\`ros2/README.md\` lines 92, 98, 99, 120, 617, 661). Those are documentation drift, not build-blocking — better off in a follow-up doc pass.

### DIFF
--- a/ros2/Dockerfile
+++ b/ros2/Dockerfile
@@ -157,8 +157,10 @@ COPY src/mowgli_localization/package.xml  src/mowgli_localization/package.xml
 COPY src/mowgli_localization/CMakeLists.txt src/mowgli_localization/CMakeLists.txt
 COPY src/mowgli_map/package.xml           src/mowgli_map/package.xml
 COPY src/mowgli_map/CMakeLists.txt        src/mowgli_map/CMakeLists.txt
-COPY src/mowgli_nav2_plugins/package.xml  src/mowgli_nav2_plugins/package.xml
-COPY src/mowgli_nav2_plugins/CMakeLists.txt src/mowgli_nav2_plugins/CMakeLists.txt
+# mowgli_nav2_plugins (FTCController) was removed in da71b13 when the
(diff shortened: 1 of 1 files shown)


cedbossneo:
Closing — this targets `feat/opennav-coverage`, which is a dead branch (the upstream `opennav_coverage` server is `COLCON_IGNORE`'d; we use `mowgli_coverage` on F2C 2.0.0). The Dockerfile cleanup is obsolete.

PR #214, opened 2026-05-15
fix(gps): align diagnostic fix-status with /gps/fix

cedbossneo:
## Symptom
Daisy reported on 2026-05-15: \"certains messages sont publish avec des status fix/float incohérent, l'un dit fix l'autre dit float\". Verified live on the robot:

\`\`\`
/gps/fix.status.status   = 2 (STATUS_GBAS_FIX → \"RTK Fixed\")
/diagnostics \"GPS: fix\"  = \"RTK Float — converging, not yet validated\"
                              (carr_soln=float, read live from /ubx_nav_status)
\`\`\`

## Cause
\`gps_health_aggregator\` read \`/ubx_nav_status.carr_soln\` directly to build the human-readable message. The HP NavSatFix node in the ublox driver caches carr_soln on its own NAV-STATUS callback and stamps it on the next NAV-HP-POS — up to ~150 ms of lag. The two consumers race on every RTK Float↔Fixed transition.

Downstream (navsat_to_absolute_pose → /gps/absolute_pose → BT → GUI) all chain off /gps/fix, so the dashboard shows one status and the diagnostic tile shows the other.

## Fix
\`gps_health_aggregator\` now subscribes to \`/gps/fix\` and sources the human-readable message from \`NavSatFix.status.status\` (when fresh, <3 s old). The raw \`carr_soln\` / \`diff_soln\` / \`fix_type\` stay in the keyvalue block.

STATUS_SBAS_FIX is overloaded by the HP node for both RTK Float and plain DGPS, so the message disambiguates via the local carr_soln keyvalue. When /gps/fix hasn't been seen recently (driver not yet up, NMEA-only without HP), fall back to the legacy logic so nothing regresses.

## Test plan
- [ ] Compare \`/gps/fix.status.status\` and \`/diagnostics\` \"GPS: fix\" message during an RTK borderline period — they should agree
- [ ] Confirm dashboard GPS badge matches the diagnostic tile across Float ↔ Fixed transitions
- [ ] No-RTK / NMEA-only smoke test: legacy fallback path still produces a reasonable message

### DIFF
--- a/sensors/gps/gps_health_aggregator.py
+++ b/sensors/gps/gps_health_aggregator.py
@@ -29,6 +29,7 @@ from rclpy.qos import QoSProfile, ReliabilityPolicy, DurabilityPolicy
 
 from diagnostic_msgs.msg import DiagnosticArray, DiagnosticStatus, KeyValue
 from rtcm_msgs.msg import Message as RtcmMessage
+from sensor_msgs.msg import NavSatFix, NavSatStatus
 
 try:
     from ublox_ubx_msgs.msg import UBXNavStatus, UBXNavSat, UBXRxmRTCM, UBXNavCov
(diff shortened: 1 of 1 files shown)


cedbossneo:
Closing — this predates and is superseded by the Universal GNSS integration (#366, merged) and the C++ topic-bridge port, which reworked GNSS backend/diagnostic handling. The branch also now conflicts with `dev`. Thanks for the work here — much of the intent landed via #366; please re-open a fresh PR against `dev` if something specific is still missing.

PR #220, opened 2026-05-16
fix(bt): show "Mowing Complete" instead of "Coverage Failed" on clean exit

cedbossneo:
## Symptom

Users on certain garden layouts (typically small lawns) report "Coverage Failed" in the GUI immediately after the 2 perimeter rings finish — before any inner mowing starts — even though the robot actually mowed the entire area.

User words: > "Systématiquement après les deux tours périmétriques, entre la fin et le début de la tonte intérieure"

## Root cause

`GetNextUnmowedArea` in `coverage_nodes.cpp` returns `BT::NodeStatus::FAILURE` in two cases the BT couldn't tell apart:

1. **Clean exit** — every area was processed; `get_coverage_status` reports `strips_remaining=0` (rings stamped `mow_progress` fully, no inner needed). On a small lawn, 2 perimeter rings cover the whole area.
2. **Real error** — service unavailable, timeout, no areas defined.

`StripGuards` returns FAILURE in both cases → `StripCoverageWithRecovery` falls to `FailedCoverageDock` → `PublishHighLevelStatus(COVERAGE_FAILED_DOCKING)` fires. Banner says "Coverage Failed" even on successful runs.

## Fix

Lift the discriminator into the BT layer:

- New `BTContext::mowing_completed_normally` flag. Raised on every clean-exit branch of `GetNextUnmowedArea`; left false on real errors (service down, timeout, "no areas defined"). Reset at the top of `onStart` (before the service-ready check) so a prior session's true value can't leak into a fresh run.
- New `IsMowingComplete` BT condition node. Tiny — just reads the flag.
- `main_tree.xml`: wrap the existing `FailedCoverageDock` in a `MowingExitChooser` Fallback whose **first branch** is guarded by `IsMowingComplete` and publishes `MOWING_COMPLETE_DOCKING` then docks + `IDLE_DOCKED` + `EndSession` + `ClearCommand`. **Second branch** is the unchanged `FailedCoverageDock` sequence — byte-for-byte identical to before, so genuine failures still surface as red banner.
- `EndSession` also resets the flag (belt + braces, for sessions that end via rain-dock or LOW_BATTERY without re-entering `GetNextUnmowedArea`).
- GUI: `MOWING_COMPLETE_DOCKING` gets a friendly label (\"Mowing Complete — Returning to Dock\"), a `tone: success` mapping in the dashboard registry, and entry in `MowerStatus.tsx`'s `MOTION_STATES` so the colour is right.

## Test plan

- [x] `colcon build --packages-select mowgli_behavior --merge-install` — green inside the `mowgli-ros2:main` image
- [x] `tsc && vite build` — green
- [ ] On the robot: mow a small area where 2 perimeter rings cover everything → banner should read "Mowing Complete — Returning to Dock" / success tone, not "Coverage Failed"
- [ ] Negative-path probe: start a mowing run with `map_server_node` killed → `get_coverage_status` times out → banner should still read "Coverage Failed" (existing path unchanged)
- [ ] Negative-path probe: start with zero recorded areas → first `get_coverage_status` returns `success=false` with `areas_queried=0` → still "Coverage Failed" (we leave `mowing_completed_normally=false` here because it really is a setup error)

## Notes

- The fix is layered ENTIRELY at the BT exit decision — it does not change what triggers `AreaLoop` to exit. Coverage planning, FTC following, and the existing 5-retry FollowStrip loop are untouched.
- If the underlying behaviour is in fact "F2C failed to plan inner sub-cells on a tiny field" (rather than "2 rings really cover everything"), the area still flows through `AreaUnreachable → AlwaysSuccess → AreaLoop advance` exactly as before — and on the final exit, this PR picks the right banner. Either way, the surface lie is fixed.


cedbossneo:
Closing: superseded on dev — the `coverage_all_complete` flag + `IsCoverageComplete` routing to the MOWING_COMPLETE terminal (instead of COVERAGE_FAILED_DOCKING) are now part of the tree (bt_context.hpp / main_tree.xml). If any of the GUI status-string polish from this branch is still wanted, it can be re-proposed as a small standalone PR.

PR #247, opened 2026-05-18
fix(bt): low-pass filter v_battery before deriving battery_percent

danyial:
## Symptom

A normal mowing run terminated in `LOW_BATTERY_DOCKING` while the GUI was still rendering ~35 % battery remaining. The BT fired correctly: `battery_percent` really had crossed 20 % — because the motors were heavily loaded and the rail had sagged 0.5–1 V from PWM-startup / stiction transients. As soon as the BT reacted (`SetMowerEnabled false` / `StopMoving`) the current draw collapsed, `v_battery` recovered to ~25.6 V, and the GUI snapped back to ~35 %.

The trigger was numerically correct but operationally premature — it acted on the under-load voltage, not on the true SoC.

## Cause

`power_sub_` in `behavior_tree_node.cpp` fed `msg->v_battery` straight into the linear voltage→percent interpolation, with no smoothing. Every sub-second sag showed up as a percent dip, and every percent-gated BT condition (`NeedsDocking` at `battery_low_percent`, `IsBatteryLow` at `battery_critical_percent`, `PreFlightCheck`) saw it.

## Fix

Low-pass the rail voltage before deriving the percent, with a **time constant** (`kBatteryFilterTauS = 2.0 s`), not a fixed EWMA weight.

The distinction matters: `/hardware_bridge/power` is published once per firmware status packet (`hardware_bridge_node::handle_status`), and the firmware emits those every `STATUS_NBT_TIME_MS = 250 ms` — **4 Hz**, while the sim's `fake_hardware_bridge_node` runs at 10 Hz. A fixed alpha silently encodes one of those rates and behaves differently on the other (and drifts again the day anyone retunes the packet period). Deriving `alpha = dt / (tau + dt)` per sample pins the behaviour to `TAU_S` regardless of rate, and makes the knob something you can size against a transient's duration.

Also in this PR:

- **An invalid reading can no longer become a 0 % gauge.** `update()` returns `nullopt` until a valid sample (≥ 10 V) has arrived, and the caller leaves `battery_percent` alone rather than clamping a glitch reading to `v_empty`.
- **`IsBatteryLow`'s redundant voltage gate now reads the filtered voltage** (`BTContext::battery_voltage_filtered`) instead of the raw rail. It is inert at the default (`battery_critical_voltage = 0.0` disables it), but any operator who set it would otherwise have re-opened the exact false trip this PR closes, from the other half of the same node. ⚠️ **Safety-relevant:** a pack that is genuinely below the threshold still trips, one time constant (~2 s) later.
- **The logic is extracted and unit-tested.** `battery_filter.{hpp,cpp}` is pure — no ROS — following the same shape as `localization_health.hpp` / `status_snapshot.hpp`.

The 20 % docking threshold and every other percent-based BT condition are unchanged; this only adjusts what voltage they see. No new ROS parameters.

## Tests

`test_battery_filter.cpp`, 13 cases:

| Property | Test |
|---|---|
| A 600 ms sag at 25.6 V does not cross the 20 % trip, and recovers | `SubSecondMotorSagDoesNotCrossTheDockingThreshold` |
| A genuine discharge still trips (lags, does not block) | `GenuineDischargeStillReachesTheThreshold` |
| 4 Hz robot and 10 Hz sim smooth identically per wall-clock second | `SmoothingIsTheSameOnTheRobotAndInSim` |
| ~63 % of a step in one tau | `ReachesMostOfAStepAfterOneTimeConstant` |
| No value — and so no percent write — until a valid sample | `NoValueUntilAValidSampleArrives` |
| Bootstrap does not ramp up from 0 V | `BootstrapsOnTheFirstValidSampleWithoutRampingUp` |
| Dropouts hold the last good value, then snap to the fresh reading | `HoldsTheLastGoodValueThroughADropout`, `SnapsToTheFreshReadingAfterALongDropout` |
| Stalled `/clock` (dt = 0) does not divide by zero or freeze wrong | `StalledClockLeavesTheValueUnchanged` |
| `tau <= 0` disables smoothing; `reset()` re-bootstraps | `NonPositiveTauDisablesSmoothing`, `ResetRequiresANewBootstrap` |
| Percent interpolation clamps, and survives a degenerate range | `InterpolatesAndClampsToTheEndpoints`, `DegenerateRangeReportsZeroRatherThanDividingByZero` |

Verified locally on `ros:kilted` (`colcon build` + `colcon test`): **291 tests, 0 failures, 44 skipped**, lint clean.

## Field test plan (still owed)

- [ ] Mowing run with intermittent heavy load (blade engagement, stiction recovery): compare raw `/hardware_bridge/power.v_battery` against the derived `battery_percent` — percent should stay monotone within the noise of the actual discharge.
- [ ] `LOW_BATTERY_DOCKING` fires only when the *unloaded* rail is genuinely near 24.8 V, not during a transient dip.
- [ ] Cold start with no `/hardware_bridge/power` messages yet → no spurious 0 % battery until the first valid sample arrives.

## Notes

Rebased onto `dev` (was targeting `main`) and revised after review; @danyial's original commit is preserved as the first commit.

### DIFF
--- a/ros2/src/mowgli_behavior/CMakeLists.txt
+++ b/ros2/src/mowgli_behavior/CMakeLists.txt
@@ -42,6 +42,7 @@ find_package(rcl_interfaces REQUIRED)
 
 add_executable(behavior_tree_node
   src/behavior_tree_node.cpp
+  src/battery_filter.cpp
   src/condition_nodes.cpp
   src/navigation_nodes.cpp
   src/docking_nodes.cpp
@@ -515,6 +516,24 @@ if(BUILD_TESTING)
     tf2
     tf2_ros
     tf2_geometry_msgs
+  )
+
+  # ---- test_battery_filter ----
+  # Regression for the premature LOW_BATTERY_DOCKING trip: battery_percent used
+  # to be interpolated off the RAW rail voltage, so a sub-second motor sag
+  # (PWM startup / stiction, 0.5-1 V) crossed battery_low_percent at a genuine
+  # 30-40 % SoC. Pins that a brief sag does not trip while a real discharge
+  # still does, that the smoothing is rate-independent (4 Hz robot vs 10 Hz
+  # sim), and that an invalid reading never becomes a 0 % gauge.
+  # Pure logic — no ROS, so no ament_target_dependencies needed.
+  ament_add_gtest(test_battery_filter
+    test/test_battery_filter.cpp
+    src/battery_filter.cpp
+  )
+
+  target_include_directories(test_battery_filter PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include>
   )
 
 endif()
--- a/ros2/src/mowgli_behavior/include/mowgli_behavior/battery_filter.hpp
+++ b/ros2/src/mowgli_behavior/include/mowgli_behavior/battery_filter.hpp
@@ -1,0 +1,108 @@
+// Copyright 2026 Mowgli Project
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: GPL-3.0
+
+#ifndef MOWGLI_BEHAVIOR__BATTERY_FILTER_HPP_
+#define MOWGLI_BEHAVIOR__BATTERY_FILTER_HPP_
+
+#include <optional>
+
+namespace mowgli_behavior
+{
+
+// ── Why this exists ────────────────────────────────────────────────────────
+// battery_percent is a linear interpolation of v_battery between the empty and
+// full endpoints, and every percent-gated BT condition (NeedsDocking at
+// battery_low_percent, IsBatteryLow at battery_critical_percent, PreFlightCheck)
+// reads it. Feeding the raw rail voltage in means motor transients decide when
+// the robot docks: PWM startup and stiction-overcome pull the pack down
+// 0.5–1 V for well under a second. A run reported from the field ended in
+// LOW_BATTERY_DOCKING while the GUI still showed ~35 %, because the sag briefly
+// crossed the 20 % voltage — and the moment the BT reacted (SetMowerEnabled
+// false / StopMoving) the load collapsed, the rail recovered to ~25.6 V, and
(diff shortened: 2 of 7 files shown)


danyial:
The red CI checks here are pre-existing on upstream `main`, not regressions from this PR:

- **Formatting (clang-format)**: complaints are about `ros2/src/fusion_graph/src/fusion_graph_node.cpp` and `ros2/src/mowgli_simulation/src/fake_hardware_bridge_node.cpp` — neither of which this PR touches. Both files last changed in `a314e82f` (`feat(fusion_graph): robustness pass`) and `364cad30` (`feat(sim): migrate to Webots`) respectively. The clang-format job globs `find ros2/src/ -name "*.cpp"` so any unrelated formatting drift on `main` fails every PR's check.
- **Build / Build & Test**: failing on either a fork-PR `ghcr.io` push-permission ("denied: installation not allowed to Write organization package") or a `rosdep` resolution for `opennav_coverage_msgs` / `opennav_coverage`. Neither is in this diff.
- **claude-review / label / review / suggest**: org-secrets aren't exposed to fork PRs.

The single check that runs on the changed file (`Static Analysis (cppcheck)`) **passes**. Happy to rebase / patch if any of this needs follow-up on my end.

whcrg:
I feel this would be very good idea, is there some problem/design issue why not merged?

cedbossneo:
Picked this up — thanks for the patch and for the CI triage, @danyial. The premise was right; I've retargeted to `dev` (repo convention) and pushed a revision on top of your commit rather than replacing it.

Four things came out of review:

1. **The rate assumption was off.** `alpha=0.05` was sized for "~10 Hz", but `/hardware_bridge/power` is published once per firmware status packet and `STATUS_NBT_TIME_MS = 250` → **4 Hz**. That gave the robot tau ≈ 4.75 s instead of the intended 2 s, while the sim (`fake_hardware_bridge_node`, 10 Hz) got ≈ 2 s. Alpha is now derived per sample from elapsed time, `alpha = dt / (tau + dt)`, so the behaviour is pinned to the time constant at any rate.
2. **The invalid-sample fallback still produced 0 %.** When the filter was un-bootstrapped and the sample was < 10 V, `v_for_pct = v_raw` clamped to `v_empty` → 0 %, which is the opposite of what the third test-plan bullet promised. `update()` now returns `nullopt` until a valid sample arrives and the caller leaves `battery_percent` alone.
3. **`IsBatteryLow`'s voltage gate was left on the raw rail.** Inert by default (`battery_critical_voltage = 0.0`), but it would have re-opened the same false trip for anyone who set it — from the other half of the same node. It reads the filtered voltage now.
4. **Tests.** The logic moved out of the subscription lambda into `battery_filter.{hpp,cpp}` (pure, no ROS — same shape as `localization_health.hpp`), with 13 gtests. Full suite: 291 tests, 0 failures, lint clean on `ros:kilted`.

On CI: `dev` has since replaced the whole-tree clang-format glob with a changed-lines `git-clang-format` check, so that class of unrelated red is gone. Verified locally against clang-format-18 — the changed lines are clean.

Field validation is still owed; the plan is in the description.

cedbossneo (approved):
CI green across the board (Build & Test ROS2 kilted, runtime image amd64 + arm64, clang-format, cppcheck, config drift). Locally verified on ros:kilted: 291 tests, 0 failures, 13 new. Merging to dev for field testing — the field-test plan in the description is still owed before dev → main.

PR #280, opened 2026-06-05
perf(map_server): cut idle CPU — gate expensive rebuilds, cache OccupancyGrids, fix reachability oscillation

yannlossouarn:
## Summary

Reduces `map_server_node` idle CPU from ~43% → ~5% on RPi4 (measured docked, 5 mowing areas defined).

Three root causes were found and fixed:

### 1. Expensive data-topic recomputation ran every timer tick

`coverage_cells_to_occupancy_grid()` (O(cells × polygons) point-in-polygon sweep) and `GridMapRosConverter::toMessage()` (4-layer serialization) were called unconditionally every 1 Hz publish tick regardless of whether any layer had changed.

**Fix:** gate both behind `content_dirty_ || masks_dirty_`. `content_dirty_` is set by:
- `mark_cells_mowed()` — mow progress stamped
- `apply_decay()` — but only when `prog.maxCoeff() > 0` (new early-out skips the full Eigen pass on an empty grid)
- `on_occupancy_grid()` — but only when an OCCUPANCY cell value actually changes (new per-cell comparison)

### 2. GUI map goes blank after foxglove page reload while idle

foxglove_bridge subscribes with volatile QoS and does not relay `transient_local` latches to WebSocket clients. With publish-on-change the map would disappear after a reload whenever the robot was idle.

**Fix:** cache the recomputed `OccupancyGrid`s in `cached_mow_progress_` / `cached_coverage_cells_` members. Recompute inside the dirty block (expensive, gated). Publish from cache every tick (cheap, unconditional). `~/mow_progress` and `~/coverage_cells` use volatile QoS; `~/grid_map` keeps `transient_local` for Nav2 `costmap_filter`.

### 3. Reachability BFS oscillated on overlapping area boundaries → `masks_dirty_` every 2 s

When two mowing area polygons share boundary cells, each area's periodic BFS disagreed: area A marked the shared cells DEAD (its seed can't reach them through an obstacle), area B marked them LAWN (its seed can). This set `masks_dirty_` on every 2 s cycle, triggering a keepout-mask republish, a KeepoutFilter reload in Nav2's global costmap, **and** a full `coverage_cells_to_occupancy_grid()` rebuild — completely bypassing the dirty-flag gate above.

**Fix:** in `recompute_reachability_for_area()`, before flipping a cell DEAD, check whether it also belongs to another mowing area. If so, skip it — another area's BFS can reach it, so DEAD would be semantically wrong and will just oscillate.

### 4. Missing `dock_approach_overshoot` default (bonus fix)

`navigation.launch.py` read `dock_approach_overshoot` before assigning its default, causing a Python `NameError` when `mowgli_robot.yaml` is absent (CI / fresh installs).

## Test plan

- [ ] Docked idle robot with 5 mowing areas: confirm `map_server_node` CPU stays <10% sustained
- [ ] Confirm no `reachability(area N): M → DEAD` log lines repeat on every 2 s tick
- [ ] Reload foxglove GUI while idle — `~/mow_progress` and `~/coverage_cells` remain visible (not blank)
- [ ] Start autonomous mowing — `mark_cells_mowed` sets `content_dirty_`, grid updates correctly
- [ ] Fresh install without `mowgli_robot.yaml` — navigation launch does not raise `NameError`


yannlossouarn:
**Validation run — 2026-06-06, RPi4B 4GB, 5 mowing areas, robot docked idle**

Rebuilt `mowgli_map` in-container from this branch and measured steady-state CPU (sampled 4× at 30 s intervals, system fully settled):

| Process | Before (old binary) | After (this PR) |
|---------|---------------------|-----------------|
| `map_server_node` | ~48% | **~9.5%** |
| `fusion_graph_node` | ~25% | ~6% (10 Hz via #281) |
| `lifecycle_manager_navigation` | ~21% | ~22% (unchanged) |

`map_server_node` CPU is stable and flat across samples — confirms the oscillation has stopped. No repeated `reachability … → DEAD` log lines observed.

Note: PR body says "~43% → ~5%" — actual measured drop is 48% → 9.5%, so the reduction is ~5×. Still very significant; the difference from the estimate is likely due to boundary conditions (number of shared polygon vertices, publish rate).

cedbossneo:
Superseded by #286 — a merge-simulation against dev is empty (the map_server idle-CPU caching + reachability fix from this PR are already on dev via the #285/#286 stack). Closing to avoid a redundant merge.

Issue #191, opened 2026-05-09
Wire mowgli_robot.yaml.undock_speed to BT BackUp speed (currently divergent: 0.15 vs 0.20)


cedbossneo:
**Source:** Nav2 + mowgli_robot.yaml audit, 2026-05-10 (commit `08e2ea55` follow-up).

## Issue

`mowgli_robot.yaml` declares:
\`\`\`yaml
undock_speed: 0.15  # m/s
\`\`\`

But `ros2/src/mowgli_behavior/trees/main_tree.xml:271,370` hardcodes:
\`\`\`xml
<BackUp backup_speed="0.20" backup_dist="1.5" .../>
\`\`\`

The GUI exposes `undock_speed` as a tunable knob, but changing it has no effect on the runtime — the BT XML wins. Result: the GUI lies to the operator.

## Acceptance

Either:

1. **Wire it through.** Use a BT input port + xacro-style substitution in `main_tree.xml` so the YAML value reaches the `BackUp` action node at launch. Same pattern as the `dock_pose_x/y/yaw` injection in `navigation.launch.py`.
2. **Or remove `undock_speed` from `mowgli_robot.yaml`** and remove the GUI control. Document the canonical 0.20 m/s in the BT XML comment.

`undock_distance: 1.5` happens to match the XML — same wiring or same removal.

## Files

- `ros2/src/mowgli_bringup/config/mowgli_robot.yaml:181-182`
- `ros2/src/mowgli_behavior/trees/main_tree.xml:271,370`
- `gui/web/src/components/.../UndockingSection.tsx` (if removing)

cedbossneo:
Done in dev via commit 9cec91a3. behavior_tree_node declares an undock_speed parameter, pushes it onto the BT blackboard, and the three undock-flow BackUp instances in main_tree.xml now use backup_speed="{undock_speed}". full_system.launch.py reads undock_speed from mowgli_robot.yaml and passes it through. OBSTACLE_BACKOFF stays hardcoded (recovery, not undock).

PR #284, opened 2026-06-06
fix(behavior): wire undock_distance param to BT blackboard

yannlossouarn:
## Problem

`undock_distance` in `mowgli_robot.yaml` was silently ignored. The three BackUp nodes in the undock flows (initial undock, rain-resume undock, battery-resume undock) all used hardcoded distances (1.0 m and 1.5 m respectively), so changing `undock_distance` in the YAML had no effect on robot behaviour.

`undock_speed` was already wired correctly (issue #191). `undock_distance` was the same pattern but never finished.

## Root cause

`undock_distance` was not declared as a ROS parameter in `behavior_tree_node.cpp` and not injected from `robot_params` in `full_system.launch.py`, so the BT blackboard never received it and the XML attributes `backup_dist="1.0"` / `"1.5"` were never parameterised.

## Fix

Exact same pattern as `undock_speed` (#191):

1. **`full_system.launch.py`** — inject `undock_distance` from `robot_params` alongside `undock_speed`
2. **`behavior_tree_node.cpp`** — `declare_parameter<double>("undock_distance", 1.0)` + `blackboard_->set("undock_distance", ...)`
3. **`main_tree.xml`** — replace `backup_dist="1.0"` / `"1.5"` with `backup_dist="{undock_distance}"` in the three undock-flow BackUp nodes; the obstacle-wedge recovery BackUp (`0.40 m`) stays hardcoded — it is not an undocking maneuver
4. **`mowgli_robot.yaml`** — default updated from 1.5 m → 1.0 m (the 1.5 m figure was the old resume-undock literal; 1.0 m is a more representative default); `CalibrateHeadingFromUndock` constraint (must be > 0.5 m) noted in the comment

## Test plan

- [ ] Set `undock_distance: 0.8` in `mowgli_robot.yaml`, start mowing — robot should reverse exactly 0.8 m on initial undock
- [ ] Same for rain-resume and battery-resume undock paths
- [ ] Obstacle-wedge recovery BackUp still reverses 0.40 m (unaffected)
- [ ] `CalibrateHeadingFromUndock` succeeds with `undock_distance` > 0.5 m; fails and retries as expected if set below 0.5 m

### DIFF
--- a/ros2/src/mowgli_behavior/src/behavior_tree_node.cpp
+++ b/ros2/src/mowgli_behavior/src/behavior_tree_node.cpp
@@ -421,16 +421,17 @@ private:
     blackboard_->set("dock_pose", dock_pose);
     blackboard_->set("undock_pose", undock_pose);
 
-    // Undock reverse speed, sourced from mowgli_robot.yaml.undock_speed
-    // and consumed by the BackUp BT instances in main_tree.xml via the
-    // {undock_speed} blackboard reference. Previously the speed was
-    // hardcoded as a string attribute in three undock-flow BackUps,
-    // which kept the yaml value orphan — editing the GUI slider did
-    // nothing. Recovery-side BackUps (e.g. OBSTACLE_BACKOFF) intentionally
-    // stay hardcoded; they are not "undocking" so they should not move
-    // when the operator tunes undock speed. See issue #191.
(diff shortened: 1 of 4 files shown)


cedbossneo:
Superseded by #286 (merged to dev). The undock_distance BT wiring from this PR is included verbatim in #286.

PR #285, opened 2026-06-07
fix(gui): mobile map performance, action visibility, and UI positioning fixes

yannlossouarn:
## Summary

- **Polygon rendering on mobile**: `DrawControl` sync timer increased from 0 ms to 300 ms — prevents a race against MapboxGL GL context initialization after `mapKey` remount, which was causing the sync to be cancelled before it ran. Stabilized `drawableFeatures` memo with a geometry-keyed ref so pose WebSocket updates (up to 50 Hz) no longer invalidate the DrawControl's feature array on every frame.
- **Map performance on mobile**: Pose stream updates throttled via `requestAnimationFrame` (one `setState` per animation frame instead of one per WS message). LiDAR stream disabled on mobile viewports. Coverage canvas `toDataURL()` deferred via rAF.
- **Manual Mowing visibility**: Added Manual Mow button directly into both the mobile toolbar (`MapToolbarMobile`) and the desktop toolbar (`MapToolbar`) — previously it was only accessible inside the "More ⋯" dropdown.
- **Joystick positioning**: `JoystickOverlay` accepts a `bottomOffset` prop; on mobile it is set to 172 px to render above the toolbar.
- **Mobile toolbar positioning**: Toolbar `bottom` raised to 108 px so it clears the browser's native navigation bar on Chrome/Firefox mobile.
- **Manual mode state on load**: `useManualMode` now accepts `stateName` and syncs `manualMode` from `MANUAL_MOWING` robot state — previously the button showed the wrong state if the robot was already in manual mode when the page loaded.
- **Cache-Control headers**: `index.html` served with `no-cache, no-store, must-revalidate` so mobile browsers always fetch the latest build.

## Test plan

- [ ] Open GUI in Chrome Mobile and Firefox Mobile — map polygons should render on load
- [ ] Confirm map is smooth while robot publishes pose (no perceptible lag)
- [ ] Confirm Manual Mowing button is directly visible in the toolbar (not in dropdown)
- [ ] Activate Manual Mowing — joystick appears above the toolbar, not behind it
- [ ] Reload page while robot is in MANUAL_MOWING state — button shows "Stop Manual" immediately
- [ ] Hard-reload after a new deploy — browser fetches fresh `index.html`

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -1,5 +1,14 @@
 # Mowgli ROS2 v3 — Environment configuration
 # Copy to .env and edit: cp .env.example .env
+# Then manage the stack with: ./docker/stack.sh up | down | logs -f | regen
+
+# Compose project name — drives the named-volume prefix (install_mowgli_maps).
+# Keep stable across the life of an install; renaming orphans persisted maps.
+COMPOSE_PROJECT_NAME=install
+
+# Optional fragments toggled by docker/stack.sh (off = mowgli + gps + gui only)
+ENABLE_MQTT=false
+ENABLE_WATCHTOWER=false
 
 # ROS2 Domain ID — all containers must share the same value for DDS discovery
 ROS_DOMAIN_ID=0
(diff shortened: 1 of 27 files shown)


cedbossneo:
Superseded by #286 (merged to dev). #286 contains every line of this branch (#286 ⊇ #285 ⊇ #284); GitHub won't auto-close it because the work was re-committed under new SHAs.

PR #290, opened 2026-06-10
feat(nav2): v2 FTC control cascade — FTC transit, derivative-LPF anti-hunt, dual Hybrid-A* planners

yannlossouarn:
## Summary
Ports the ROS1 **v2 Follow-The-Carrot control cascade** into the ROS2 stack to fix flaky in-place rotation, transit "hunting", and imprecise dock targeting. FTC now drives transit + coverage + dock approach (single smooth-reference carrot → ω → near-open-loop motor layer), mirroring the v2 stack that maneuvered reliably.

Rebased/merged onto current `dev` (incorporates #278/#281/#283/#286; parallel TF-lead + stack.sh work reconciled).

## Key changes
- **`FollowPath` = `FTCController`** (was RotationShim+RPP) with v2 transit gains — RPP couldn't fine-pivot this robot's static-friction-deadband motors. (invariant #8)
- **Derivative low-pass filter** on FTC's PID derivative (`derivative_filter_tau`, opt-in): the raw finite-difference derivative differentiated noisy 10 Hz fused-pose feedback, and `kd_lat` pumped it into a ~1.5 Hz steering limit cycle. Filtering kills the hunt without sacrificing cross-track tracking.
- **Dual Hybrid-A\* planners**: `GridBased` (DUBIN, forward-only) for transit/coverage-start — no Reeds-Shepp reverse-cusp at goals; `GridBasedRS` (REEDS_SHEPP) used only by the dock approach, which needs reverse to set up the approach in the tight, obstacle-adjacent dock area.
- **`DockApproach`** BT node: Smac(`GridBasedRS`)→staging + straight dock-aligned tail → one FTC `FollowPath`, replacing the oscillating opennav graceful-controller arc.
- Relaxed waypoint yaw a touch (FTC `max_goal_angle_error` 12→15°, `stopped_goal_checker` yaw 0.25→0.30 rad; invariant preserved).

## Validated on the robot this session (area "TestArea")
- **Transit hunting fixed**: LPF on vs off on a real Smac path → wz-command chatter −31%, gyro chatter −30%, cross-track −40% (50 vs 83 mm). Transit on Dubins paths ~0.09 ω-reversals/s (essentially hunt-free).
- **Full TestArea coverage** ran cleanly: localization solid (σ 8–36 mm RTK throughout), smooth swaths.

## Status — DRAFT (docking approach is WIP)
Transit + coverage are solid. The **dock approach is not finished**:
- the return transit can leave the robot off the dock approach axis;
- the lethal block by the dock trips FTC collision-abort → replan/hunt;
- the momentum seat can trip the firmware EMERGENCY on hard contact (needs a gentler/slower final approach).
These are the subject of a dedicated follow-up session. Opening as draft for visibility/review of the transit/coverage work.


cedbossneo:
Great work ! I'm waiting a little bit before merging because i'm also challenging mppi for coverage as we need obstacle avoidance

PR #292, opened 2026-06-14
Feat/universal gnss integration

Pepeuch:
## Summary

Replaces the legacy receiver-specific GNSS stack with the new Universal GNSS integration.

This PR removes the old u-blox / Unicore / NMEA-specific drivers, containers and installer paths, then migrates MowgliNext to a unified `sensors/gps` runtime powered by Universal GNSS. It also adds the GUI and backend workflow needed to configure, apply, reset and monitor GNSS receivers from MowgliNext.

## Changes

* Removed legacy GNSS sidecars and receiver-specific runtime paths:

  * old `sensors/unicore`
  * old `sensors/nmea`
  * old u-blox / NMEA / Unicore compose files and install hooks
  * legacy GNSS status adapter code
* Added Universal GNSS as the unified GNSS integration path.
* Added unified `sensors/gps` sidecar with Universal GNSS runtime.
* Added Universal GNSS topic bridge for MowgliNext ROS2 topics.
* Updated ROS2 launch/config files to consume Universal GNSS status and RTCM topics.
* Updated installer, compose files and Docker workflows for the new GNSS runtime.
* Added vendor-neutral GNSS configuration in the GUI:

  * Receiver Profile
  * Signal Profile
  * Position Rate
  * Runtime Baud
  * Configured Receiver Baud
  * Expert receiver settings
* Added GNSS backend API for:

  * planning receiver configuration
  * applying receiver profile
  * restarting GPS sidecar
  * factory reset + profile apply
* Added live GNSS status/health display in the GUI.
* Updated documentation for Universal GNSS migration and GUI design.
* Updated CI/CD workflows for the unified GPS sidecar and removed legacy receiver-specific workflows.

## Component

* [x] ROS2 Stack (`ros2/`)
* [x] GUI (`gui/`)
* [x] Docker (`docker/`)
* [x] Sensors (`sensors/`)
* [ ] Firmware (`firmware/`)
* [x] Documentation (`docs/`)
* [x] CI/CD (`.github/`)

## Testing

* [x] Built successfully
* [x] Tested on hardware
* [ ] Tested in simulation
* [x] Unit tests pass
* [x] Manual testing

Hardware validation performed with:

* Unicore UM982
* serial transport at 460800 baud
* NTRIP corrections
* RTCM forwarding
* RTK Fixed state
* healthy Universal GNSS parser diagnostics
* working GUI/API plan/apply/restart flow

Observed on robot:

* Universal GNSS status published at 5 Hz
* valid GNSS fix
* RTK corrections active
* RTCM frames forwarded to the receiver
* clean serial stream at 460800 baud
* receiver configuration planning with `--signal-profile`
* GUI and backend able to drive the Universal GNSS configuration workflow

## Checklist

* [x] Follows conventional commit format
* [x] Documentation updated (if user-facing)
* [x] No hardcoded secrets or credentials
* [x] No unrelated changes

## Notes

This is an intentional migration PR. The legacy GNSS sidecars are removed and replaced by the unified Universal GNSS runtime.

Known follow-ups:

* Add per-constellation satellite statistics to the GUI.
* Improve RTK mode labels in the GUI.
* Expose advanced signal/C/N0 diagnostics.
* Validate u-blox hardware path.
* Validate generic NMEA hardware path.
* Add future developer-mode raw receiver command tools.


gui/pkg/api/settings.go
```diff
@@ -308,6 +309,298 @@ func coerceValue(value string, schemaType string) any {
...
+		"GNSS_PROFILE":         profile,
+		"GNSS_SIGNAL_PROFILE":  signalProfile,
+		"GNSS_PROFILE_RATE_HZ": profileRateHz,
+		"GNSS_BACKEND":         "universal",
+		"GNSS_NTRIP_ENABLED":   boolStringValue(flat["ntrip_enabled"], true),
+		"GNSS_NTRIP_HOST":      stringValue(flat["ntrip_host"], "crtk.net"),
+		"GNSS_NTRIP_PORT":      stringValue(flat["ntrip_port"], "2101"),
+		"GNSS_NTRIP_MOUNTPOINT": ntripMountpoint,
+		"GNSS_NTRIP_USERNAME":   stringValue(flat["ntrip_user"], "centipede"),
+		"GNSS_NTRIP_PASSWORD":   stringValue(flat["ntrip_password"], "centipede"),
```
cedbossneo, line 476: **High / security.** `stringValue` treats `""` the same as unset (`settings.go:317`), and the schema ships `ntrip_password`/`ntrip_user` defaults as `""` (`mower_config.schema.json:325`). So a user who *deliberately clears* their NTRIP credentials gets `GNSS_NTRIP_USERNAME/PASSWORD=centipede` (and `host=crtk.net`) written into the runtime `.env` on every settings save and every baud apply (`settings.go:1056`, `gnss.go:593`). That silently re-enables a login to the public Centipede caster the operator intended to disable, and the YAML keeps the empty value so it's invisible. Distinguish "unset" from "explicitly empty" before falling back to a default.

install/compose/docker-compose.gps.yml
```diff
@@ -34,16 +34,20 @@ services:
...
-      GPS_DEVICE_PATH: ${GPS_PORT:-${UBLOX_DEVICE_SERIAL_STRING:-/dev/gps}}
-      GPS_PROTOCOL: ${GPS_PROTOCOL:-UBX}
-      GPS_BAUD: ${GPS_BAUD:-921600}
+      GNSS_STACK: ${GNSS_STACK:-universal}
+      GNSS_RECEIVER_FAMILY: ${GNSS_RECEIVER_FAMILY:-auto}
+      GNSS_TRANSPORT: ${GNSS_TRANSPORT:-serial}
+      GNSS_SERIAL_DEVICE: ${GNSS_SERIAL_DEVICE:-/dev/gps}
+      GNSS_SERIAL_BAUD: ${GNSS_SERIAL_BAUD:-921600}
+      GNSS_FRAME_ID: ${GNSS_FRAME_ID:-gps_link}
+      GNSS_NTRIP_ENABLED: ${GNSS_NTRIP_ENABLED:-true}
```
cedbossneo, line 43: **High.** Every GNSS var here uses a default substitution (`${GNSS_NTRIP_ENABLED:-true}`, `${GNSS_SERIAL_DEVICE:-/dev/gps}`, …), so Compose always passes a non-empty value into the container. The sidecar resolvers in `sensors/gps/start_gps.sh` check the env var *first* and only fall back to `parse_yaml`, which makes the `mowgli_robot.yaml` keys (`gnss_serial_device`, `ntrip_enabled: false`, …) **dead** in the real deployment — and a compose default can override an explicit YAML value (YAML `ntrip_enabled: false` still runs NTRIP because the env default is `true`). Either drop the env defaults so YAML wins, or document that `.env`/GUI is the single source of truth and hand-editing YAML has no effect.

sensors/gps/start_gps.sh
```diff
@@ -23,158 +21,248 @@ parse_yaml() {
...
+  yaml_serial_baud="$(parse_yaml gnss_serial_baud)"
+  if [ -n "$yaml_serial_baud" ]; then
+    printf '%s\n' "$yaml_serial_baud"
+    return 0
+  fi
+
+  printf '921600\n'
+}
+
+resolve_ntrip_enabled() {
```
cedbossneo, line 91: Resolver precedence is env-first → YAML-fallback. Combined with the Compose `${VAR:-default}` substitution in `docker-compose.gps.yml`, the YAML fallback branch is unreachable in the containerized deployment (see the compose comment). Worth aligning on one source of truth.

install/lib/gps.sh
```diff
@@ -64,296 +64,107 @@ preset_key_loaded() {
...
+  case "$receiver_family" in
+    ublox|auto) probe_backend="ublox" ;;
+    unicore) probe_backend="unicore" ;;
+    *)          probe_backend="nmea" ;;
+  esac
+
+  if [[ "$baud_preconfigured" != "true" && -n "$probe_port" ]]; then
+    prompt_or_probe_baud "$probe_port" "$probe_backend" "$default_baud" "$probe_mode"
+    GNSS_SERIAL_BAUD="$REPLY"
+    maybe_upgrade_unicore_baud "$probe_port" "$GNSS_SERIAL_BAUD" "$probe_mode"
```
cedbossneo, line 159: **High.** The default receiver family is now `auto`, but `maybe_upgrade_unicore_baud` (`unicore_config.sh:83`) returns early unless the family is literally `unicore`, so a Unicore UM98x on the default `auto` never gets its 921600 upgrade. Worse, `ublox_backend_selected` treats `auto` as ublox, so `maybe_upgrade_ublox_baud` (line 160) then runs, fails UBX MON-VER detection on the Unicore unit, and **clamps `GNSS_SERIAL_BAUD` to the low probed baud** (`ublox_config.sh:94`). Probe ordering at line 151 also uses the ublox sweep for all `auto` receivers. The vendor baud-upgrade guards need to handle `auto` (detect-then-upgrade) rather than require the explicit family the old flow set.

install/lib/config.sh
```diff
@@ -362,48 +712,129 @@ parse_args() {
       --lang=*)
         MOWGLI_LANG="${1#*=}"
         ;;
-      --branch=*|--channel=*|--image-tag=*)
+      --branch=*)
```
cedbossneo, line 715: **Medium / behavior change.** `--branch=` and `--image-tag=`/`--channel=` were split; `--branch=` now only sets `REPO_BRANCH` and `IMAGE_TAG` defaults to `main`. The decoupling is intentional (`mowglinext.sh:113`), but it silently changes behavior: existing non-interactive deploys using `--branch=dev` now check out the dev git branch yet pull `:main` containers (previously `--branch=dev` was an `IMAGE_TAG` alias and pulled `:dev`). Worth a release note / migration mention.

gui/pkg/api/gnss.go
```diff
@@ -0,0 +1,657 @@
...
+	if err != nil {
+		return err
+	}
+
+	if baudInt, convErr := strconv.Atoi(runtimeBaud); convErr == nil {
+		doc.Flat["gnss_serial_baud"] = baudInt
+	} else {
+		doc.Flat["gnss_serial_baud"] = runtimeBaud
+	}
+	gnssEnvUpdates := applyUniversalGnssCompatibility(doc.Flat)
```
cedbossneo, line 579: **Medium.** `persistGNSSRuntimeBaud` is named/messaged as a narrow baud write, but it calls `applyUniversalGnssCompatibility`, which regenerates the *entire* GNSS env block (NTRIP host/user/pass/profile/device) from the YAML flat and rewrites `.env`. Combined with the `centipede` default coercion in `settings.go:476`, applying just a baud change can clobber `.env` NTRIP creds with the public defaults. Either scope this to baud only, or rename/document the wider side effect.

sensors/gps/start_gps.sh
```diff
@@ -23,158 +21,248 @@ parse_yaml() {
   grep -E "^\s+${1}:" "$CONFIG" | head -1 | sed 's/.*:\s*//' | tr -d '"' | tr -d "'"
```
cedbossneo, line 21: **Medium (latent).** `sed 's/.*:\s*//'` is greedy and `tr -d` strips all quotes, so a YAML value containing `:` (an NTRIP password like `ab:cd`, or a `host:port`) is truncated at the last colon. Latent because env normally wins, but it breaks any standalone-sidecar run that relies on YAML creds. Consider `sed 's/^[^:]*:[[:space:]]*//'` and quote-aware handling.

sensors/gps/universal_gnss_topic_bridge.py
```diff
@@ -0,0 +1,204 @@
...
+            rtcm_qos,
+        )
+
+        self.get_logger().info(
+            "Bridging Universal GNSS topics: "
+            f"{input_status_topic} -> {output_status_topic}, "
+            f"{input_rtcm_topic} -> {output_rtcm_topic}"
+        )
+
+    def _on_status(self, msg: UniversalGnssStatus) -> None:
```
cedbossneo, line 143: **Medium.** The old `navsat_to_absolute_pose_node` republished `/gps/status` on every `/gps/fix` (5+ Hz), so status liveness tracked position liveness. The bridge now publishes `/gps/status` only when the receiver emits on the internal status topic. If that internal status stream stalls while `/gps/fix` keeps flowing, the GUI shows a stale RTK state (e.g. last-known RTK_FIXED) after corrections have actually dropped. Consider a staleness guard or tying status liveness to fix arrivals.

install/lib/config.sh
```diff
@@ -305,33 +594,94 @@ is_supported_gnss_backend() {
...
+}
+
+gnss_serial_device_from_state() {
+  if [[ -n "${GNSS_SERIAL_DEVICE:-}" ]]; then
+    printf '%s\n' "$GNSS_SERIAL_DEVICE"
+    return 0
+  fi
+
+  case "$(gnss_connection_from_serial_device)" in
+    usb)  printf '/dev/serial/by-id/usb-stub\n' ;;
```
cedbossneo, line 669: **Medium.** A non-interactive `--gnss-connection=usb` with no device returns this `/dev/serial/by-id/usb-stub` placeholder, which is then persisted into both `.env` and `mowgli_robot.yaml`. `check_generated_gps_yaml_alignment` passes because both sides hold the same fake value, masking the misconfiguration, and the sidecar then tries to open a nonexistent device. Fail the check (or refuse to persist) on this sentinel instead of treating it as aligned.

cedbossneo (commented):
Automated correctness + cleanup review of the Universal GNSS integration. Core logic (navsat node math, bridge mapping, QoS contracts) looks intact; the inline comments below flag credential/security handling, the env-vs-YAML precedence model, an installer baud regression for Unicore on the new `auto` default, and a few cleanups. Severity is noted per comment.

universal-gnss-runtime-only.patch
```diff
@@ -0,0 +1,3153 @@
+diff --git a/gui/pkg/api/settings.go b/gui/pkg/api/settings.go
```
cedbossneo, line 1: **Cleanup.** This 3153-line raw `.patch` file (and `universal-gnss-docs-entrypoints.patch`) is a byte-for-byte duplicate of edits already present in this PR, referenced nowhere — a workflow artifact accidentally committed to the repo root. Please delete both (and consider adding `*.patch` to `.gitignore`).

gui/pkg/providers/transform.go
```diff
@@ -64,48 +64,131 @@ type rawTwistWithCovariance struct {
...
-		return 5
-	case 0:
-		return 1 // plain fix — only the FLAG_GPS_RTK base bit, NOT FIXED
+	case 0, 1, 2:
+		return 1
 	default:
 		return 0
 	}
+
+	return 0
```
cedbossneo, line 191: **Cleanup.** Unreachable `return 0` — the switch's `default` already returns 0, so `go vet` reports `transform.go:191:2: unreachable code` (breaks a vet-gated CI stage). Also: the doc comment above `navSatStatusToFlags` still describes the old RTK 3/5 flag mapping, which now contradicts the collapsed-to-1 logic from `f4c2ff4d` — update or remove it.

PR #293, opened 2026-06-14
fix(gnss): address Universal GNSS review follow-ups

cedbossneo:
Follow-up fixes for the review of #292 (Universal GNSS integration). Scoped to the **clear, low-risk** findings; the deferred ones are listed below with rationale.

## Fixed

| Sev | Area | Fix |
|-----|------|-----|
| High | NTRIP creds → public caster | Only fall back to the Centipede login (`centipede/centipede`) when the receiver is actually pointed at `crtk.net`. A custom caster no longer receives those credentials, and a deliberately-cleared credential is no longer silently restored. Applied in all three layers that injected it: `gnssCompatFromFlat` (GUI), `env.sh` (installer), `start_gps.sh` (sidecar). |
| High | env vs YAML precedence | `docker-compose.gps.yml` passes `GNSS_*` through **without** hard defaults, so an explicit `mowgli_robot.yaml` value (e.g. `ntrip_enabled:false`, a custom serial device) is no longer silently overridden by a compose default. The sidecar resolvers carry the previous effective defaults (`/dev/gps`, NTRIP-on, etc.) as last resort, so the fully-unconfigured behavior is unchanged. |
| Med | baud apply side effect | `persistGNSSRuntimeBaud` now writes **only** the baud keys to the runtime `.env` instead of regenerating the entire GNSS env block — a baud-only apply can no longer clobber NTRIP/device values. |
| Med | `parse_yaml` truncation | Strips only the first `key:` prefix and a single pair of surrounding quotes, so values containing `:` (NTRIP passwords, `host:port`) survive. Verified against `p@ss:w0rd:42`. |
| Low | dead code / `go vet` | Removed the unreachable `return 0` in `navSatStatusToFlags` and refreshed its stale RTK-flag doc comment. |
| Low | repo hygiene | Deleted the two raw `.patch` files (3.9k lines) accidentally committed to the repo root. |

## Deferred (with rationale)

- **Unicore baud upgrade skipped on `family=auto`** — this is **intentional**, guarded by `test_unicore_config.sh:78-82` ("generic universal backend does not send Unicore commands"). `auto` deliberately avoids sending vendor commands to an unidentified receiver; the supported path is selecting `family=unicore`. Making `auto` auto-detect-and-upgrade is a product decision, not a clear bug fix.
- **`--branch=dev` pulls `:main` images** — intentional decoupling of git branch from image channel (`mowglinext.sh:113`). Documenting/warning is the right follow-up, not reverting.
- **`/dev/serial/by-id/usb-stub` placeholder persisted** — `usb-stub` is a deliberate sentinel baked into the test harness (`harness.sh`, `test_gps_matrix.sh`, `test_env_output.sh`); changing it is invasive and out of scope here.
- **`/gps/status` staleness** (bridge only republishes on internal status, not on `/gps/fix` rate) — a real medium, but the safe fix touches safety-relevant status display and wants hardware validation. Tracking separately.

## Testing

- `go build ./...`, `go vet ./pkg/...` clean.
- Go: `TestGNSS*`, `TestPostSettingsYAML*`, `TestApplyUniversalGnssCompatibility*` pass. (Two pre-existing failures — `TestGetSettingsSchema_FromUpstream` (network) and `TestBuildBoard` (empty DB path) — fail identically on `dev` and are unrelated.)
- Installer: `test_env_output` (86), `test_compose_validity` (28), `test_robot_yaml` (19), `test_state_parsing` (35), `test_gps_matrix` (18), `test_unicore_config` (8), `test_ublox_config` (15), `test_negative` (21) — all pass.


Pepeuch:
Thanks, this follow-up addresses the blocking review points for #292.

I agree with keeping the remaining items as follow-ups:
- auto receiver baud upgrade should be handled as a dedicated detection workflow, not guessed during install;
- image channel / branch decoupling needs documentation;
- GPS status staleness and SATINFO/NAV-SAT GUI should be handled in a dedicated diagnostics/UI pass.

I tested the Universal GNSS chain on hardware:
- UM982: RTK Fixed with NTRIP and GUI status working.
- u-blox F9P: persistent apply now succeeds, CFG-RAM+BBR+FLASH is used, and RTK Fixed is reported through `/gps/status`.

Remaining GUI work after this PR:
- improve satellite/constellation diagnostics display;
- reduce/inspect minor u-blox malformed parser warnings;
- prefer RTK mode over plain fix type in the GPS card.

Pepeuch (approved):
Reviewed and tested on hardware.

The changes address the review comments and do not introduce any regression in the Universal GNSS workflow.

Approved.

PR #305, opened 2026-06-20
fix(nav2): stop transit-pivot oscillation (pivot once, then pursue)

wjcloudy:
## Summary
On a transit (`FollowPath` = RotationShim + RegulatedPurePursuit), the in-place rotate-to-heading **re-engages mid-transit against the frozen transit plan** and oscillates / runs to a wrong heading, so the transit never completes. The same in-place-rotate path is shared with coverage swath pivots.

Hardware-confirmed on a Yardforce 500B (ROS 2 Kilted, prebuilt images): with this change the robot pivots once at the start, drives onto the path, and starts mowing - no oscillation. Without it, the robot just sits at the start of the path turning left and right.

## Root cause
`RotationShimController::getSampledPathPt()` samples its rotate target **0.5 m from `current_path_.poses.front()`** - the plan **START** - assuming the plan begins at the robot. That holds only for a fresh plan. The transit plan is **never replanned from the current pose** (observed `plan_age` climbing 0.7 s -> 27 s in one run), so once the robot drives off `poses.front()`, the re-engaged in-place rotate (RPP's `use_rotate_to_heading` + the shim re-triggering on each `setPlan`) aims at that stale point and runs away.

The **initial** pivot always worked (plan fresh, robot ~= `poses.front()`); only the re-engaged rotate on the stale plan failed. RPP **pursuit** prunes the plan to the robot and steers correctly - it was never the problem.

## Fix - pivot once, then pursue
| Param | Before | After | Why |
|---|---|---|---|
| `rotate_to_heading_once` | (absent -> false) | **true** | shim pivots only on the first command per plan |
| `use_rotate_to_heading` | true | **false** | RPP pursues mid-path bends instead of re-rotating in place |
| `closed_loop` | true | **false** | open-loop ramp, required on this deadband chassis - same lesson already documented for `FollowCoveragePath` |

`rotate_to_goal_heading` is **kept `true`**: the terminal goal pivot uses the fresh goal pose (not the stale plan), so it is not affected by this bug, and it is required for `opennav_docking` terminal heading alignment.

## Testing
Hardware-confirmed 2026-06-20 (Yardforce 500B, deadband chassis, single-hall encoders, MPU6050). Heading-error trace (computed from `/plan` + TF `map->base_link`, not the fused pose): shim pivots -115 deg -> -14 deg (deadband cleared), disengages, RPP drives the path (cmd_vx ~0.4 m/s), mowing starts. No oscillation across repeated runs.

## Notes for review
- The mid-path-bend hand-off that `use_rotate_to_heading: true` provided is given up; on a mower following smooth transit paths, RPP pursuit handles bends without an in-place rotate. If a deployment has very sharp mid-transit bends, this is the param to revisit.
- A robust upstream fix would have `getSampledPathPt()` prune to the robot's nearest point on the path (as RPP does) instead of `poses.front()`, so the start pivot is correct even on a stale plan; the param change here avoids the failure without touching nav2.

### DIFF
--- a/ros2/src/mowgli_bringup/config/nav2_params_base.yaml
+++ b/ros2/src/mowgli_bringup/config/nav2_params_base.yaml
@@ -211,6 +211,14 @@ controller_server:
       rotate_to_heading_angular_vel: 0.75
       max_angular_accel: 2.5
       simulate_ahead_time: 1.0
+      # Pivot ONCE at the start of each transit, then hand to the primary
+      # controller — do NOT re-engage the in-place rotate mid-path. The transit
+      # plan is FROZEN (never replanned from the current pose) and RotationShim
+      # samples its target 0.5 m from poses.front() (the plan START). Once the
+      # robot drives off poses.front(), a re-engaged shim/RPP rotate aims at that
+      # STALE point and oscillates / runs to a wrong heading and never aligns
+      # (root cause hardware-confirmed 2026-06-20; see use_rotate_to_heading below).
+      rotate_to_heading_once: true
       # Terminal pivot. With this false, neither RotationShim (start-of-path
       # only) nor RPP (mid-path, rotate_to_heading_min_angle: 0.785 rad =
       # 45°) does anything when the navigation ends with a yaw error in the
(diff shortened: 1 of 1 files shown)


wjcloudy:
There are still issues even with this (maybe caused by it) as it sometimes seems to go off path and get stuck in circles during mow. But this patch at least lets it drive the initial transit rather then just moving left and right.

cedbossneo:
Closing: these exact changes (rotate_to_heading_once: true, use_rotate_to_heading: false, closed_loop: false on the FollowPath shim — with your root-cause comments) are already live on dev, absorbed via 966fbef9. Thank you for the hardware-confirmed root-cause work — fresh field logs today showed exactly this pathology on an older image (35 s+ sub-deadband pivot loop with 1 Hz replans), confirming the fix matters. Anyone still seeing it should pull the current :dev image.

PR #346, opened 2026-07-02
fix(map_server): boundary slack band is high-cost, not free — kills phantom corridors between areas

cedbossneo:
Field report: **the robot cuts directly from one area to another** wherever their edges run close, ignoring the operator's navigation areas ("50 cm de chaque côté d'autorisé").

**Root cause**: the outside boundary-slack band (`enforce_boundary_margin_m` = 0.25 m around **every** area, needed so RTK drift at the edge doesn't self-reject) was **FREE (0)** in the keepout mask. Two areas whose edges run within 2×margin expose a ~0.5 m free strip — a *phantom corridor* Smac happily shortcuts through instead of the navigation areas.

**Fix**: the slack band becomes **high-cost (90 → keepout cost ≈229)**:
- non-lethal → boundary RTK drift still tolerated, last-resort crossing still plannable if **no** legal route exists,
- expensive → the planner always prefers any navigation-area route.

Tests: existing keepout test updated to the new semantics + new `KeepoutMaskNoFreePhantomCorridorBetweenCloseAreas` regression (two rectangles 0.30 m apart → the gap must cost 90). 50/50 mowgli_map tests green.

Note (même rapport terrain): les "transits à travers les obstacles" visibles sur la capture sont les connecteurs lame-active du planificateur pré-#340 — corrigés par #340/#345 ; vérifié par simulation sur la carte réelle de l'utilisateur (3 zones, 7 obstacles) : **0 pose dans les obstacles** avec le planificateur actuel.

### DIFF
--- a/ros2/src/mowgli_map/src/costmap_filters.cpp
+++ b/ros2/src/mowgli_map/src/costmap_filters.cpp
@@ -30,6 +30,13 @@
 namespace mowgli_map
 {
 
+// OccupancyGrid value for the outside boundary-slack band. KeepoutFilter maps
+// mask values to cost as round(v × 254 / 100): 90 → ~229 — steep enough that
+// Smac never routes through the band when ANY legal route (navigation area)
+// exists, non-lethal so RTK drift a few cm outside the boundary does not read
+// as "in collision" and a last-resort crossing remains plannable.
+constexpr int8_t kBoundarySlackCost = 90;
+
 void MapServerNode::publish_keepout_mask()
 {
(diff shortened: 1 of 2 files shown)


cedbossneo:
Closing: superseded on `dev` — the outside slack band is already mid-cost via `kOutsideSlackMaskCost = 50` in `ros2/src/mowgli_map/src/costmap_filters.cpp` (deliberately lower than the 90 proposed here). The phantom-corridor regression test from this branch is worth cherry-picking separately.

PR #410, opened 2026-07-26
fix: pipe chassis mass into URDF

### LINKED ISSUE #193: Pipe chassis_mass_kg from mowgli_robot.yaml into mowgli.urdf.xacro (currently hardcoded)
**Source:** Nav2 + mowgli_robot.yaml audit, 2026-05-10.

## Issue

\`\`\`yaml
# mowgli_robot.yaml:27
chassis_mass_kg: 8.76
\`\`\`

\`\`\`xml
# ros2/src/mowgli_bringup/urdf/mowgli.urdf.xacro:83
<xacro:property name="base_mass" value="8.76"/>
\`\`\`

Same value, two places, no automated link. The GUI exposes `chassis_mass_kg` for editing; changing it has no effect on the URDF \(robot_state_publisher reads the xacro at boot, the YAML value never reaches it\).

## Acceptance

Same pattern already used for `chassis_length`, `chassis_width`, `chassis_height`, `chassis_center_x`, `wheel_radius`, etc. in `mowgli.launch.py:91-117`:

\`\`\`python
chassis_mass_kg = float(robot_params.get("chassis_mass_kg", 8.76))
# ... pass via xacro args
\`\`\`

Then in `mowgli.urdf.xacro`:
\`\`\`xml
<xacro:arg name="chassis_mass_kg" default="8.76"/>
<xacro:property name="base_mass" value="$(arg chassis_mass_kg)"/>
\`\`\`

## Files

- `ros2/src/mowgli_bringup/config/mowgli_robot.yaml:27`
- `ros2/src/mowgli_bringup/launch/mowgli.launch.py`
- `ros2/src/mowgli_bringup/urdf/mowgli.urdf.xacro:83`

liltaket:
Closes #193.

## What changed
- pass `chassis_mass_kg` from `mowgli_robot.yaml` through `mowgli.launch.py` to the URDF xacro
- use the new xacro argument for `base_link` inertia mass
- add a regression test that expands xacro with a non-default mass and verifies `base_link`
- synchronize stale LiDAR-overlay defaults and their regression expectations with the already field-validated robot-template defaults

## Why
The GUI-configurable chassis mass previously had no effect because the URDF hardcoded `8.76`. During validation, CI also exposed a pre-existing configuration-drift assertion: the launch already applied the template values, while the raw LiDAR overlay and its test still used the previous defaults.

## Validation
- `colcon build --packages-select mowgli_bringup`
- `colcon test --packages-select mowgli_bringup --return-code-on-test-failure` (7/7 passed)
- `colcon test-result --verbose` (64 tests, 0 failures)
- GitHub Actions rerun pending.

### DIFF
--- a/ros2/src/mowgli_bringup/CMakeLists.txt
+++ b/ros2/src/mowgli_bringup/CMakeLists.txt
@@ -69,6 +69,9 @@ if(BUILD_TESTING)
   ament_add_pytest_test(test_robot_config_util test/test_robot_config_util.py
     TIMEOUT 30
   )
+  ament_add_pytest_test(test_urdf_xacro test/test_urdf_xacro.py
+    TIMEOUT 30
+  )
 
   # Static guard for Inv 2 (TF sole ownership) — greps sources/configs, no
   # nodes spun up.
--- a/ros2/src/mowgli_bringup/config/nav2_params_lidar.yaml
+++ b/ros2/src/mowgli_bringup/config/nav2_params_lidar.yaml
@@ -245,10 +245,10 @@ collision_monitor:
       enabled: false
     PolygonSlow:
       type: "polygon"
-      # Chassis + 15 cm — slow to 30% when obstacle is close
+      # Chassis + 15 cm — slow to 70% when obstacle is close
       points: "[[0.65, 0.35], [0.65, -0.35], [-0.25, -0.35], [-0.25, 0.35]]"
       action_type: "slowdown"
-      slowdown_ratio: 0.5
+      slowdown_ratio: 0.7
       # 12 points: the dock structure sits near the polygon boundary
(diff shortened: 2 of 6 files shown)


cedbossneo:
Heads up on the base branch: this PR targets `main` directly, whereas our internal convention routes changes through `dev` first. That's expected for an external contribution — just flagging that merging this lands it straight on `main` rather than going through the usual `dev` → `main` promotion. The change itself is self-contained and CI is green, so I'm approving as-is.

cedbossneo (approved):
LGTM. Correctly single-sources the chassis mass default (8.76) across the xacro arg and launch fallback, fixing the hardcoded value that made the GUI setting a no-op. New xacro regression test is a solid guard, and the bundled nav2 drift-sync is justified (CI lockstep guard). All checks green.

PR #411, opened 2026-07-26
fix(gui): recover from corrupted Bitcask database

### LINKED ISSUE #288: mowgli-gui crash-loops after unclean shutdown — DB corruption panics NewDBProvider ("key/value size is invalid")
### Component

GUI (web interface)

### Bug Description

After a reboot / power loss, the `mowgli-gui` container can enter a crash-loop (`Restarting`), leaving the web UI unreachable (port 4006 not listening, `curl` → HTTP 000). The container panics on startup because its embedded KV store file is corrupted.

This is recoverable but bricks the UI until the operator knows the trick — and a robot losing power uncleanly is a normal occurrence.

### Steps to Reproduce

1. Reboot the Pi / cut power while the GUI is running (unclean shutdown, mid-write).
2. `docker ps` → `mowgli-gui: Restarting (2)`.
3. Web UI on `:4006` unreachable.

### Relevant Logs

```
panic: key/value size is invalid

goroutine 1 [running]:
github.com/cedbossneo/mowglinext/pkg/providers.NewDBProvider()
	/app/pkg/providers/db.go:104 +0x94
main.main()
	/app/main.go:19 +0x2c
```

Not OOM (1.6 GiB free at the time). The store file `docker/config/db/000000000.data` was corrupted by an interrupted write during shutdown.

### Workaround that worked

```
docker stop mowgli-gui
mv docker/config/db/000000000.data{,.corrupt.bak}   # back up the corrupted file
docker start mowgli-gui                              # recreates a fresh DB -> GUI returns HTTP 200 in ~3 s
```

No important data lost: this KV store only holds GUI-side state (onboarding-completed flag, mowing session history, schedule, map offset). Dock pose, IMU/mag calibrations and mowing areas live elsewhere (`mowgli_robot.yaml`, `/ros2_ws/maps/areas.dat`, calibration files) and are preserved.

### Suggested fix

Make `NewDBProvider()` (db.go:104) resilient to a corrupted store: attempt recovery, or move the corrupt file aside and recreate a fresh store, instead of panicking. That way an unclean shutdown doesn't make the UI unreachable until manual intervention.

### Hardware & Environment

```
MowgliNext HEAD 85d157a5 (2026-06-04), images ghcr.io/cedbossneo/mowglinext/*:main
Raspberry Pi 4B (4 GB), Debian 13
```

liltaket:
## Summary

An interrupted Bitcask write can leave the GUI unable to rebuild its index, causing `NewDBProvider()` to panic and the container to restart indefinitely. This adds a narrow backup-first recovery path for Bitcask's documented truncated/corrupt-entry failures while retaining the current behavior for healthy databases and unrelated operational errors.

## Root cause

`NewDBProvider()` previously panicked on every `bitcask.Open()` error. With Bitcask v1.0.2, an interrupted final write can make index reconstruction return `data is truncated` or `key/value size is invalid`, so each GUI restart failed at the same point.

## Changes

- Retry only Bitcask's known corrupt-entry error messages.
- Copy the complete database directory before recovery, excluding the transient lock file.
- Retry with Bitcask v1.0.2's `WithAutoRecovery(true)` and log the normal-open failure, backup path, and successful recovery.
- Add regression coverage for healthy startup, truncated-data recovery, non-corruption failures, repeat startup, and backup-name collisions.

## Recovery behavior

1. Open normally.
2. If the error is `data is truncated` or `key/value size is invalid`, create a sibling backup named `<db-path>.corrupt-<UTC timestamp>-<random>`.
3. Reopen with Bitcask auto-recovery, which retains the valid prefix of the final data file and rebuilds the index.
4. If backup or recovery fails, preserve the error and panic as before; no empty database is created.
5. Permission, path, lock, disk, configuration, metadata, and other operational errors are not treated as recoverable corruption.

## Data safety

The original database directory, including all `.data` files and metadata, is copied before recovery can truncate the final corrupt record. The transient `lock` file is deliberately excluded. Existing backups are never removed or overwritten; collision-resistant names are created atomically beside the database.

## Tests

All commands below passed:

```text
cd gui
go test ./pkg/providers -run 'TestDBProvider' -count=1
go test ./pkg/providers -run 'TestDBProvider' -count=10
go test -race ./pkg/providers -count=1
go test ./... -count=1
go vet ./...
go build ./...
git diff --check
```

## Manual reproduction

Outside the repository, I created a valid temporary Bitcask store, wrote a value, closed it, marked its index stale, and truncated the final `.data` file by one byte. Provider startup logged the normal open failure, created a sibling backup, completed auto-recovery, and accepted a subsequent write. A later clean startup did not create another backup.

## Risks

This relies on the pinned Bitcask v1.0.2 recovery implementation, which only repairs the last data file and may discard its corrupt tail. The complete pre-recovery store remains in the backup for manual inspection.

## Not tested

- Real Raspberry Pi power loss.
- Disk-full and permission-denied failures during backup or recovery.
- Every historical Bitcask corruption pattern.

## Rollback

Reverting this PR restores the previous startup behavior. Existing recovery backups are not automatically deleted.

Fixes #288

### DIFF
--- a/gui/pkg/providers/db.go
+++ b/gui/pkg/providers/db.go
@@ -2,9 +2,16 @@ package providers
 
 import (
 	"errors"
+	"fmt"
 	"git.mills.io/prologic/bitcask"
 	"golang.org/x/xerrors"
+	"io"
+	"log"
 	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"time"
 )
 
 type DBProvider struct {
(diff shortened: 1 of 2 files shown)


cedbossneo (approved):
Approved. Well-scoped, well-tested reliability fix — backup-before-recovery ordering is correct, recovery is gated to the two documented bitcask corruption errors, and healthy-DB behavior is unchanged. CI green on amd64+arm64; verified gofmt/vet clean and TestDBProvider passing under -race locally.

Issue #395, opened 2026-07-23
UM980 GPS Issues


fwelvering:
I'm using a UM980 for GPS support on the MowgliNext Setup.
I'm using the settings as described in the openmower documnentation  https://openmower.de/latest/docs/step-by-step/2-robot-modification/prepare-the-parts/prepare-the-gps/ (clicked Withmotion TAB)

Getting a lot of:

universal_gnss_ntrip/rtcm.1230_not_valid
The latest RTCM 1230 GLONASS code-phase bias message is not marked valid

universal_gnss_ntrip/rtcm_semantic/glonass_code_phase_bias
RTCM semantic observation decoded but not valid

universal_gnss_ntrip/rtcm.required_messages_missing
Required RTCM correction messages have not been observed

universal_gnss/rtcm_semantic/glonass_code_phase_bias
RTCM semantic observation decoded but not valid

Takes about 2 hours to get a fix, and it loses it after a short while.

cedbossneo:
Thanks for the report. These `rtcm.1230_not_valid` / `glonass_code_phase_bias` warnings are a **known issue** with the universal-gnss NTRIP diagnostic path — the RTCM 1230 (GLONASS code-phase bias) message from many casters isn't flagged `valid` in the way the current validity check expects, so it raises false alarms even when RTK is working fine. It's being addressed in our universal-gnss fork (bridge re-pin).

A couple of things that would help confirm your setup is actually healthy despite the noise:
- Do you reach **RTK-Fixed** (check `/gps/pose_cov` σ dropping to ~mm, or the GUI GPS status)? If yes, the corrections are flowing and these are cosmetic warnings.
- Is your NTRIP mountpoint GPS-only or does it include GLONASS? A GPS+GLONASS mountpoint that omits a valid 1230 will trigger `required_messages_missing`.

If you're *not* getting a fix, that's the real problem to chase — please share your receiver model config and NTRIP mountpoint. Keeping this open to track the false-alarm suppression.

cedbossneo:
Thanks for the report — there are actually **two separate issues** here, and it helps to split them.

### 1. The RTCM 1230 / `required_messages_missing` / `glonass_code_phase_bias` alarms — false positive, fixed

[...]

PR #421, opened 2026-07-30
fix(gnss): UM980 RTK/DGPS correction-age windows + never forward a truncated RTCM frame (#395)

cedbossneo:
## Why

Follow-up on #395. `MODE ROVER UAV` (merged) did **not** fix @fwelvering's UM980: still no RTK-Fixed, now also reproduced against a *public* caster, with **every** Universal GNSS diagnostic green — `ntrip_streaming` OK, `rtcm_forwarding_active` OK, `receiver_rtcm_active` OK, all `msm_*_msm7` decoded+valid, `write_error_count 0`, `receiver_correction_available true`, `correction_age_s ~0.6`, 21/33 sats used. Yet it sits in Float for hours — while **socat-piping the same caster into the same receiver reaches Fixed in seconds**.

That points at receiver *configuration*, not the correction path. Comparing our `rover_high_precision` profile against the field-proven configs:

| command | OpenMower[1] | Centipede UM980[2] | UM980 default[3] | before | after |
|---|---|---|---|---|---|
| `MODE` | `ROVER UAV` | `ROVER UAV` | — | `ROVER UAV` | unchanged |
| `CONFIG RTK TIMEOUT` | *not sent* | 180 | 120 | **10** | **120** |
| `CONFIG DGPS TIMEOUT` | *not sent* | 300 | 300 | **600** | **300** |
| `CONFIG RTK RELIABILITY` | *not sent* | 4 3 | 3 1 | 3 1 | unchanged |

[1] `ClemensElflein/openmower.de` — prepare-the-gps · [2] `CentipedeRTK/docs-centipedeRTK` UM980 rover CONFIG · [3] dumped `CONFIG` response in `jclark/satpulse`

A 10 s RTK window ages out base observations — and the RTK filter state with them — on any correction gap longer than 10 s, forcing a full ambiguity re-convergence. The paired 600 s DGPS window then lets the receiver sit in a DGPS solution for up to ten minutes instead of pushing back toward RTK. Together that is the reported "loses Fixed, then hours to reacquire" signature.

## Changes

`universal-gnss` re-pinned to `a6d3cf3` (branch `fix/rover-dynamic-mode-uav`, `mowglinext/universal-gnss`):

- `rover_high_precision` now emits `CONFIG RTK TIMEOUT 120` / `CONFIG DGPS TIMEOUT 300`, as named constants carrying the field evidence in the comment.
- `CONFIG RTK RELIABILITY 3 1` kept — it *is* the receiver default; explicit so a hand-tightened receiver gets the known-good value back.
- Docs updated (`docs/vendors/unicore/config_profiles.md`).
- Drive-by compile fix: `test_receiver_auto_config.cpp` was missing the `ReceiverAutoConfigRoverDynamicMode` using-declaration and **did not compile** after the `MODE ROVER UAV` change.

No parent-repo code changes — the GUI exposes `gnss_unicore_rtk_timeout_s` / `gnss_unicore_dgps_timeout_s` as optional expert overrides with no baked defaults, so both fall through to the profile builder.

## Also in this PR: never forward a truncated RTCM frame

Second submodule commit (`4bf5d0a`), same `#395` investigation.

The RTCM forward path wrote to a **non-blocking** fd and treated `{bytes_written = 0, status = kOk}` — the POSIX `EAGAIN` case, i.e. a momentarily full kernel TX buffer — as a reason to stop. Whatever prefix had already gone out stayed on the wire, so the receiver saw a **truncated RTCM3 frame** and failed its CRC. The only symptom was `write_error_count` incrementing with a `none` transport error, which reads like a benign hiccup rather than corrupted corrections.

- New `universal_gnss_transport::WriteFrame` resumes across short writes and waits out transient stalls with a bounded, **progress-resetting** retry budget (default ~64 ms per stall — enough to drain a 4 KiB TX buffer at 115200 bps), so a frame either lands whole or is reported as damaged.
- The write policy lives in `gnss_transport`, not the ROS node, so it is unit-testable without a serial port or a ROS graph.
- Two previously-conflated failures are now distinct in the `universal_gnss/rtcm_forwarding` diagnostic:
  - `truncated_frame_count` — a prefix reached the wire and the frame could not be finished, so the peer is looking at a **corrupt** frame.
  - `stall_retry_count` — zero-progress attempts that were waited out, i.e. the link is at or near saturation even when frames complete.

Note this is *not* what's biting @fwelvering (their `write_error_count` is 0) — it's a latent corruption path found during the investigation.

## Test plan

Built and ran the affected suites locally (plain CMake sources, clang++):

- [x] `test_unicore_config_profile_builder` — pass. New assertions pin `RTK TIMEOUT 120`, `RTK RELIABILITY 3 1`, `DGPS TIMEOUT 300`; verified **RED** first (both new assertions fail against `10`/`600`).
- [x] `test_receiver_auto_config` — pass (did not compile before this branch).
- [x] `test_receiver_config_application`, `test_unicore_response_router` — pass.
- [x] `test_frame_writer` (new, 8 cases) — pass; verified **RED** against the old break-on-stall behaviour (4 failures). Covers: complete write, short-write resume, stall-then-complete, budget-resets-on-progress, exhausted budget ⇒ `truncated`, stall-before-any-byte ⇒ *not* truncated, hard error propagation, empty/null inputs.
- [x] `test_transport_foundation` — pass (no regression).
- [ ] `gnss_ros2/receiver_node.cpp` is **not** locally buildable (needs ROS); relying on the main-repo colcon CI.
- [ ] **Field confirmation owed** — @fwelvering to report whether time-to-Fixed collapses from hours to seconds/minutes.

## Notes / follow-ups (not in this PR)

- `gnss_unicore_rtk_timeout_s` and `gnss_unicore_dgps_timeout_s` are **dead settings**: rendered and persisted, but never forwarded to `gnss_config_apply` by either `gui/pkg/api/gnss.go` or `sensors/gps/start_gps.sh`. An operator override silently does nothing.
- Every GPS container start re-applies the profile (`sensors/gps/start_gps.sh:507`), and `MODE ROVER …` resets the receiver's position engine — so every restart zeroes RTK convergence. Saving on the GPS settings page triggers an apply/restart (`gui/pkg/api/gnss.go:230`); a plain settings save on other pages does **not**. Making the `MODE` command idempotent needs a way to read the receiver's current mode first — design still open.
- The CRC-failed RTCM frame **count** (`invalid_rtcm_frames`) is collapsed to a boolean and never surfaced numerically in diagnostics.


anttix:
#### Also in this PR: never forward a truncated RTCM frame

I have observed RTCM frame corruption with NMEA receiver ( Quectel LC29HEA ) as well. Glad to see a fix landing.

PR #422, opened 2026-08-01
fix: one-click dock calibration end-to-end + calm docking approach

Chatoese:
## Summary

Field-debugged on a YardForce 500 on a sloped lawn (2026-08-01, 6 instrumented runs). The Settings → Docking **Start dock calibration** button did nothing; unwinding that surfaced a chain of independent defects, each verified live on hardware:

1. **Calibration never started moving** — `run_dock_calibration_core` polled `gps_rtk_fixed_` during the RTK-Fixed wait, but the feeding `/gps/absolute_pose` subscription was only created *after* the wait (and torn down after every run), so every GUI-triggered run timed out with NO_RTK while the robot sat still.
2. **No feedback in the GUI** — `topicSubscribeInterval()` did not know `dockCalibrationStatus`, so MultiplexRoute dropped the subscription and neither progress nor failure messages reached the browser (log: `ignoring unknown topic`).
3. **COG gates physically impossible** — per-sample COG scatter is set by the publisher's 0.10 m baseline vs 2–3 cm RTK noise (field-measured 10–27° std on straight 2 m reverses); the 3° std ceiling could never pass, and the 6° bearing gate tripped on the mean's own 3–4° standard error.
4. **Dock yaw source** — the persisted yaw is now the reversed RTK travel bearing (~1° over 2 m) instead of the noisy COG circular mean; the COG mean is demoted to a coarse cross-check (`dock_cog_gate.hpp` + tests updated).
5. **Re-dock drive delegated to the production docking pipeline** (HL_CMD_HOME). Three custom forward controllers were field-tested and each found a new failure mode (blind dead-reckoning missed on slope slip; point-bearing P corrected late/steep; Stanley on the fused yaw inherited a startup-transient bias that also blinded its own angle guard). The calibration now only supervises: an angle guard on the raw RTK track course aborts an askew final approach, then a steered ~2 m backoff re-lines the robot for a fresh attempt (max 3). Persist retries ride out map_server's yaw-convergence gate.
6. **Docking controller calmed** — `v_angular_max` 0.45 → 0.25 and `k_phi/k_delta` 5.0/2.0 → 3.0/1.2: the 1/r curvature blow-up still saturated ω in the last 30 cm (~45° over-rotation, left-right limit-cycle wobble, outer wheel digging in on soft ground — same signature as the 2026-06-25 log).

## Field verification

With all fixes + a re-captured dock pose: staging transit, straight final approach, contact on the first attempt, stable 28 V charge, `fusion_graph` re-anchored at the dock.

## Notes for review

- The dock-yaw source change (travel bearing vs COG mean) revises 'Resolution A' — the gate header documents the reasoning; unit tests updated and still pin NO +pi.
- Template `mowgli_robot.yaml` gate defaults changed in lockstep with the code defaults (Invariant 15: defaults live in the template).
- Safety: blade stays off throughout calibration; firmware remains sole blade/e-stop authority; the re-dock leg now uses the same supervised production pipeline as normal HOME.


Chatoese:
Note for reviewers: commit fe3d274 bumps the universal-gnss submodule 7343753 → a6d3cf3. The pinned 7343753 does not compile on its own (missing using-declaration in its test, upstream dev CI red since 2026-07-30); a6d3cf3 fixes that and carries the issue-#395 RTK/DGPS timeout fix. Please review the bump consciously — it touches GNSS receiver configuration.

cedbossneo (approved):

PR #425, opened 2026-08-06
fix(gui): render power-menu confirm via App.useApp().modal, not static Modal

whcrg:
## Summary

The header power menu (Restart Mowgli / Restart board / Restart Pi / Shut down Pi)
did nothing when clicked — no confirm dialog, no request, nothing in logs.

## Root cause

Every item is gated behind `confirmAction()`, which used the **static**
`Modal.confirm`. On this stack (React 19 + antd v5, no
`@ant-design/v5-patch-for-react-19`) the static Modal API renders outside the
`<App>`/`ConfigProvider` tree and the confirm dialog never appears, so `onOk`
never fires and no `/system/*` or `/containers/*` request is ever sent. The
backend routes (`system.go`, `containers.go`) and handlers were fine all along.

## Fix

Pull `modal` from `App.useApp()` (as `notification` in this file already was)
and call `modal.confirm(...)`. This is the same pattern every other component
uses (`MowerActions`, `FlashBoardComponent`, `DriveMotorSection`,
`HardwareSection`, `GnssReceiverActionsCard`, `RobotComponentEditor`);
`MowerStatus` was the lone straggler still on the static API.

## Component
- [x] GUI (`gui/`)

## Testing
- [x] Built successfully
- [x] Tested on hardware

## Checklist
- [x] No unrelated changes


whcrg:
Actually not needed as this is included in https://github.com/mowglinext/mowglinext/pull/426

Issue #430, opened 2026-08-11
GUI: investigate and reduce high energy use


liltaket:
### Component

GUI (web interface)

### Problem

Keeping the MowgliNext GUI open on macOS can use noticeably more energy than expected during passive monitoring.

### Steps to Reproduce

1. Open the MowgliNext web GUI on macOS.
2. Leave the dashboard or map page open while the GUI receives normal live updates.
3. Observe browser CPU, GPU, memory, network traffic, and energy impact over several minutes.
4. Compare the result with an idle tab and between the dashboard, map, and logs views.

### Expected Behaviour

The monitoring UI remains responsive while keeping idle and steady-state CPU, GPU, memory, network traffic, and energy use reasonably low, including on laptops and lower-powered clients.

### Measurements

Capture repeatable before-and-after measurements for:

- CPU time and long tasks
- JavaScript heap and browser process memory
- transferred bytes, request counts, and WebSocket traffic
- render and update frequency
- browser energy impact where available

Measure both idle behaviour and representative live update rates.

### Candidate Areas

- topic update and render frequency
- duplicate requests or subscriptions
- map and layer updates
- log list rendering
- continuous animation and blur effects
- static asset compression and caching

### Progress

- [x] Static asset transfer: #432 adds precompressed gzip sidecars and immutable caching; the measured compressed transfer total is 70.77% smaller.
- [ ] Map updates: #433 limits visual pose updates to 10 Hz and lidar rendering to 5 Hz. The branch remains in draft while display-mode behaviour is defined in #443.
- [x] Log rendering: #434 batches high-rate WebSocket lines; the measured median time to render 1,000 lines fell from 10,679 ms to 5,539 ms.
- [ ] Continuous visual effects: #439 removes production backdrop-filter layers and decorative infinite animations in steady-state views. The branch remains in draft while display-mode behaviour is defined in #443.
- [x] GUI test contract: #440 aligns manual-mode tests with the current Command 7 / Command 8 and blade-off safety behavior; the full Vitest suite passes 253/253.
- [x] Schema/template parity: #441 aligns four GUI defaults with the checked-in robot template; the complete Go test, vet, and build checks pass.
- [ ] Direct macOS power telemetry remains to be captured with an available energy monitor; the completed browser measurements quantify the sources of compositing, render, transfer, and update work without mower hardware.

### Environment

- Client: macOS laptop
- Browser: Chromium-based browser
- Server: MowgliNext GUI

[...]

PR #433, opened 2026-08-11
perf(gui): stabilize map stream rendering

liltaket:
## Summary

- render tracked-obstacle layers as direct `Source` children so `react-map-gl` supplies the source correctly
- coalesce visual pose updates to the latest value at up to 10 Hz
- coalesce lidar projection and rendering to the latest frame at up to 5 Hz
- keep the raw robot pose reference current on every incoming pose frame
- retain the pose snapshot associated with each accepted lidar frame
- cancel pending visual updates when map streams stop or the page unmounts

## Performance

Deterministic synthetic stream measurements:

| Input | Existing behaviour | Updated behaviour | Difference |
| --- | ---: | ---: | ---: |
| 100-frame burst | 100 render callbacks | 2 render callbacks | -98% |
| 100 Hz pose stream | 100 render callbacks/s | 10–11 render callbacks/s | about -89% |
| 100 Hz lidar stream | 100 projection/render callbacks/s | 5–6 callbacks/s | about -94% |

The map regression previously produced repeated errors for four tracked-obstacle layers plus a React fragment warning. The same mocked map now renders with zero source/fragment errors.

## Testing

- [x] throttle unit tests: 4 passed
- [x] complete Playwright matrix: 65 passed
- [x] map rendered with datum, pose, work area, and tracked obstacle; screenshot inspected
- [x] TypeScript typecheck
- [x] Vite production build
- [x] Docker `build-web` target
- [x] Prettier check for new files
- [x] `git diff --check` and secret scan

## Related test maintenance

The two outdated manual-mode expectations observed in the original Vitest run are corrected in #440.

Related to #430

### DIFF
--- a/gui/web/src/components/settings/DisplayModeSection.test.tsx
+++ b/gui/web/src/components/settings/DisplayModeSection.test.tsx
@@ -1,0 +1,39 @@
+import {beforeEach, describe, expect, it} from "vitest";
+import {fireEvent, render, screen} from "@testing-library/react";
+import {ThemeProvider, useThemeMode} from "../../theme/ThemeContext.tsx";
+import {DisplayModeSection} from "./DisplayModeSection.tsx";
+
+function ModeValue() {
+    return <output data-testid="display-mode-value">{useThemeMode().displayMode}</output>;
+}
+
+describe("DisplayModeSection", () => {
+    beforeEach(() => {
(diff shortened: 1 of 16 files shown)


liltaket:
Superseded by #444, which integrates the map render budget with the selectable display modes.

PR #453, opened 2026-08-12
fix(manual): keep blade off after Manual Mowing

liltaket:
## Summary

Introduces a distinct blade-disabled Manual Drive mode for supervised repositioning.

- Adds `COMMAND_MANUAL_DRIVE = 9` and `MANUAL_DRIVING` state through the behavior tree, ROS interface, hardware bridge, GUI types, and STM32 protocol.
- Keeps Manual Drive explicit in the behavior tree and lets teleop remain active while the blade is off.
- Enforces blade-off both when the STM32 receives Manual Drive and in its motor loop, so a queued blade packet cannot create a blade-on tick in this mode.
- Makes Blade Off transition from Manual Mowing to Manual Drive rather than racing a low-level override.

## Validation

- `colcon build --packages-select mowgli_behavior`
- `ctest --test-dir build/mowgli_behavior -R test_docking_boundary_exempt --output-on-failure`
- `git diff --check`

The PlatformIO compile-only check could not complete because the fresh toolchain download stalled before compilation. No hardware or upload action was performed.


cedbossneo (commented):
Not mergeable as is — it weakens the blade stop it sets out to fix:
- "Blade Off" loses its direct, BT-independent path: on `dev`, `mow_enabled: 0` goes straight to `/hardware_bridge/mower_control`; here it becomes a mode transition only. Keep the direct `mow_enabled: 0` call **in addition** to the mode change.
- `useManualMode.ts:106` drops the `mow_enabled: 0` from `handleStopManualMode`, the second independent blade-cut trigger on the Stop button.
- It also conflicts with `dev` and overlaps `blade_gate.hpp` (already on `dev` via #461) and #558's session policy; a rebase on `dev` + #558 would remove most of the diff.

liltaket:
Parking this draft for now. The implementation is no longer a priority and has drifted from current dev. Keeping the PR and branch for history/reference in case parts are useful later.

Issue #450, opened 2026-08-12
Manual teleop can stutter when cmd_vel updates are delayed


liltaket:
## Description / observed behaviour

During continuous joystick input in Manual Mow, the mower can occasionally feel jerky: drive, briefly stop, then drive again.

## Code observations

The frontend stores the latest joystick `TwistStamped` and retransmits it every 100 ms ([useManualMode.ts](https://github.com/mowglinext/mowglinext/blob/0a32f32750d94e4b4dddd78b7a367ae8c56b7ef5/gui/web/src/pages/map/hooks/useManualMode.ts#L8-L16), [L50-L60](https://github.com/mowglinext/mowglinext/blob/0a32f32750d94e4b4dddd78b7a367ae8c56b7ef5/gui/web/src/pages/map/hooks/useManualMode.ts#L50-L60)). The STM32 hard-stops motors after a command age over 200 ms ([cpp_main.cpp](https://github.com/mowglinext/mowglinext/blob/0a32f32750d94e4b4dddd78b7a367ae8c56b7ef5/firmware/stm32/ros_usbnode/src/ros/ros_custom/cpp_main.cpp#L828-L855)). This leaves limited margin for browser scheduling stalls, rendering load, Wi-Fi/WebSocket jitter, or a delayed host-to-MCU hop.

The dedicated `cmd_vel_ws_relay` is launched and is preferred by the GUI provider when connected; it otherwise falls back to Foxglove during early startup or relay unavailability ([ros.go](https://github.com/mowglinext/mowglinext/blob/0a32f32750d94e4b4dddd78b7a367ae8c56b7ef5/gui/pkg/providers/ros.go#L541-L551)). However, the relay currently publishes once per received browser frame rather than storing the desired state and publishing at an independent cadence ([cmd_vel_ws_relay.py](https://github.com/mowglinext/mowglinext/blob/0a32f32750d94e4b4dddd78b7a367ae8c56b7ef5/ros2/src/mowgli_bringup/scripts/cmd_vel_ws_relay.py#L36-L65)). The relationship to the observed stutter still needs on-robot measurement.

[...]

PR #455, opened 2026-08-13
fix(teleop): relay bounded velocity heartbeat

liltaket:
## Summary

Makes the WebSocket cmd_vel relay resilient to delayed browser frames without allowing stale motion.

- Replays a fresh velocity command every 50 ms for at most 125 ms.
- Publishes exactly one zero command when that lease expires, then remains quiet until a new input arrives.
- Publishes zero again on a clean client disconnect.
- Retains input clamping and adds focused relay tests.

## Validation

- `ctest -R test_cmd_vel_ws_relay --output-on-failure` — passing
- `git diff --check`


cedbossneo (approved):
Safety read: the relay can extend a command by at most the 125 ms lease past the last browser frame, which stays under the STM32's 200 ms watchdog — so the firmware remains the backstop and is not bypassed — and it publishes an explicit zero once the lease expires instead of leaving the last non-zero command to the watchdog. Clamping is preserved in `twist_from_json`. The test pins both the replay-then-zero behaviour and that both timing constants stay below 0.2 s. Approving.

cedbossneo:
Pushed one maintainer commit on top of the rebase: the new `test_cmd_vel_ws_relay.py` failed on CI with `ModuleNotFoundError: No module named 'websockets'`. The relay has always imported it, but it was only ever installed by the ros2 Docker image (apt `python3-websockets`); `package.xml` never declared it, so `rosdep` on the CI runner had no reason to provide it. Added it as an `exec_depend` (the runtime truth) and a `test_depend` (what unblocks CI). No change to your code. Merging once CI is green.

cedbossneo:
Approved, rebased onto `dev`, CI green — the only thing holding this back is that it is still marked **Draft**, which blocks merge. If you consider it done, mark it *Ready for review* and it will go in (closes #450).

cedbossneo (commented):
Re-checked: 125 ms lease under the firmware 200 ms watchdog, explicit zero on expiry and on client disconnect, CI green, already approved. Only needs an *Update branch* before merge.

PR #457, opened 2026-08-13
feat(gui): add landscape manual controller

### LINKED ISSUE #452: Add a landscape dual-stick mobile controller for manual driving
## Proposal

Consider a dedicated landscape controller page for phone-based manual operation. The current interface has one 2D joystick ([JoystickOverlay.tsx](https://github.com/mowglinext/mowglinext/blob/0a32f32750d94e4b4dddd78b7a367ae8c56b7ef5/gui/web/src/pages/map/components/JoystickOverlay.tsx#L17-L107)); a focused two-thumb layout could be safer and more comfortable outdoors.

The interaction inspiration is the *kind* of task-focused manual-control surface used by Mammotion: an immediately legible mower/map context plus direct manual controls, rather than copying its branding or visual design. Mammotion's public guide also documents a map page with Manual alongside Mow/Create/Edit actions and virtual joystick use while mapping ([guide](https://support.mammotion.com/portal/en/kb/articles/how-to-create-channel)).

## Suggested landscape layout

```text
+----------------------------------------------------------------+
| Manual Drive / Manual Mow | mower state | battery | connection |
|----------------------------------------------------------------|
|                                                                |
|                 lightweight live location map                  |
|        mower position + heading + boundary/no-go context        |
|                                                                |
|----------------------------------------------------------------|
| [THROTTLE]     [Transport] [Mow] [Record]     [STEERING]       |
|  left stick    [Start / Stop] [E-STOP]         right stick      |
+----------------------------------------------------------------+
```

- **Left stick:** vertical axis only for forward/reverse throttle.
- **Right stick:** horizontal axis only for steering/yaw.
- The combined values still produce the existing `TwistStamped` manual-control command.
- **Mode/action cluster:** clearly distinguish Transport/Manual Drive, Manual Mow, and Record. Start should be contextual to the selected mode; Stop must be visually distinct and always available.
- **Emergency stop:** permanently visible, high-contrast, deliberately separated from routine controls, with the resulting state clearly shown. It must continue to use the existing safety authority rather than becoming a UI-only stop.
- **Blade state:** show an unambiguous blade-off/on state and mode; do not rely on colour alone.
- **Lightweight map:** retain mower position, heading, active boundary/no-go areas, and perhaps a simple route/recording trace. Avoid full Mapbox detail, dense overlays, or live LiDAR rendering while controlling, unless explicitly requested. This is intended to reduce rendering pressure on a phone without losing orientation.

## Interaction and safety expectations

- Large, labelled controls with at least 44 × 44 px touch targets and spacing that prevents accidental taps.
- Independent throttle and steering deadzones, optional response/expo curves, and optional turtle-speed limiter.
- Each stick springs to centre; releasing either touch immediately zeros that axis. Releasing both yields zero `TwistStamped`.
- On page blur, visibility loss, connection loss, mode change, or stale input, command zero immediately; the lower-level watchdog remains in place as the final safety backstop.
- Respect landscape safe areas and avoid gesture conflicts with browser/system navigation.
- Pressed/armed/disabled states and connection freshness should be visible; no control should look actionable when its command path is unavailable.
- A Manual Mow start should make the blade-related transition explicit, with an appropriate confirmation if maintainers consider that necessary. Manual Drive must remain blade-off.

## Possible implementation direction

One possible approach could be a separate, lightweight `ManualControllerPage` or map sub-view activated in landscape mode, reusing the existing manual-control transport and status data rather than duplicating motion logic. It may be worth giving it a dedicated low-detail subscription/rendering budget. Another solution may be preferable depending on the intended GUI architecture.

## Related work

This would pair naturally with a blade-off manual-driving mode (#451) and should account for the manual teleop reliability investigation in #450. It is also related to #449, because the controller should make the selected operational mode and resulting blade state clear rather than offering competing low-level blade commands.

Feedback on whether this interaction model fits the intended UI would be very welcome.

liltaket:
## Summary

Supersedes #456, which GitHub closed when its fork head ref was renamed during repository cleanup. This replacement keeps the same controller implementation and current commit history.

- Map **Manual Mow** opens the dedicated controller instead of retaining a legacy map overlay.
- Desktop uses a smaller combined stick plus W/S drive, A/D steering, Space-to-stop, and simultaneous W+A / W+D arcs.
- Portrait uses a larger combined stick; mobile landscape uses two thumb sticks.
- Joystick and keyboard stay disabled until the requested high-level state has been observed and held for two bridge periods, allowing the STM32 blade-safe mode to settle before velocity is sent.
- Stop, Emergency Stop, Space, key release, window blur, hidden tabs, and mode changes all send zero velocity.

Depends on #453 and #455.

Closes #452

## Validation

- `yarn tsc --noEmit`
- `yarn vitest run src/pages/ManualControllerPage.test.tsx src/pages/map/components/MapToolbar.test.tsx` — 20 passing
- Rendered responsive checks at 375×667 portrait, 667×375 landscape, and 1280×760 desktop
- `git diff --check`

### DIFF
--- a/gui/web/src/components/AppShell.tsx
+++ b/gui/web/src/components/AppShell.tsx
@@ -65,6 +65,7 @@ const PAGE_META: Record<string, {titleKey: string; subtitleKey?: string}> = {
   '/statistics':  {titleKey: 'pageMeta.statistics.title', subtitleKey: 'pageMeta.statistics.subtitle'},
   '/settings':    {titleKey: 'nav.settings',             subtitleKey: 'pageMeta.settings.subtitle'},
   '/parameters':  {titleKey: 'nav.parameters',           subtitleKey: 'pageMeta.parameters.subtitle'},
+  '/manual-control': {titleKey: 'nav.manualControl', subtitleKey: 'pageMeta.manualControl.subtitle'},
   '/logs':        {titleKey: 'nav.logs',                 subtitleKey: 'pageMeta.logs.subtitle'},
   '/onboarding':  {titleKey: 'nav.onboarding'},
 };
--- a/gui/web/src/i18n/locales/en.json
+++ b/gui/web/src/i18n/locales/en.json
@@ -13,6 +13,8 @@
     "stats": "Stats",
     "settings": "Settings",
     "parameters": "Parameters",
+    "manualControl": "Manual Control",
+    "manualShort": "Control",
     "logs": "Logs",
     "onboarding": "Onboarding",
     "more": "More",
@@ -40,6 +42,9 @@
     },
     "parameters": {
       "subtitle": "Live ROS2 tuning"
+    },
+    "manualControl": {
+      "subtitle": "Joystick driving"
     },
     "logs": {
       "subtitle": "Container output"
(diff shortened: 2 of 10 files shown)


cedbossneo (commented):
Draft feedback, two blockers:
- **No client heartbeat any more** (`:115-126`): the new page delegates it to #455, which is not merged, so the relay never sees a lease and the robot stutters/stops.
- **The zero on unmount is never transmitted**: the joy-stream effect cleanup runs after the client is torn down, so `stopForFocusLoss() → sendZero()` has nothing to publish; the PR text ("unmount … send zero velocity") is not what the code does.
- Also depends on `Command 9`, which does not exist on `dev`, so the "Transport" button reports success and does nothing.
Rebase on `dev` once #455 is in, keep the heartbeat, and add tests for unmount→zero and Emergency Stop.

liltaket:
Parking this draft for now. The implementation is no longer a priority and has drifted from current dev. Keeping the PR and branch for history/reference in case parts are useful later.

PR #498, opened 2026-08-24
fix(behavior): stop a targeted mow rolling over into the next area

cedbossneo:
## Problem

A targeted mow — the GUI's "mow this area" button (`~/start_in_area` → `COMMAND_START`) — mowed the requested area and then **rolled over into another area** instead of stopping and docking.

Field log, robot, 2026-08-24:

```
1787590666  StartInArea: received area=1
1787590684  GetNextUnmowedArea: targeted run — mowing only area 1 (single-area mode)
1787590684  GetNextUnmowedArea: area 1 selected (0 swath(s) done so far) — dispatch attempt 1/5
            ... ~56 minutes mowing area 1 ...
1787594063  GetNextUnmowedArea: area 0 selected (0 swath(s) done so far) — dispatch attempt 1/5   <-- BUG
1787594113  HighLevelControl: received command=2   (operator had to send it home)
```

Note the second selection carries **no** "targeted run" line.

## Root cause

The single-area constraint lived only in `GetNextUnmowedArea`'s own members:

* `onStart()` resets `current_area_idx_ = 0` on **every** entry;
* the clip (`current_area_idx_ = target; max_areas_ = target + 1;`) was applied only from `ctx->target_area_index`, which `onStart()` **consumed** on first use.

The BT re-enters `MowingSequence` once the targeted area completes, so that second `onStart()` had nothing left to read — the clip was gone and it iterated from area 0. Consuming the optional on the first *tick* cannot distinguish a later tick of the SAME run from a NEW run.

## Fix — single-area mode becomes session state

* **`BTContext::single_area_target`** holds the clip for the whole session. `target_area_index` stays the one-shot *request*: consuming it latches `single_area_target` and — still exactly once per request — erases that area's stale `completed_areas`/`attempted_areas` flags, so an explicit re-mow of an already-mown area is not skipped. Keeping the erase tied to the request is load-bearing: repeating it on every `onStart()` would wipe the completion the targeted area just earned and re-mow it forever.
* **`onStart()` honours `single_area_target` on every entry.** The follow-up entry now finds the target already completed, the skip loop overshoots `max_areas_`, and the node sets `coverage_all_complete` + returns `FAILURE`.
* **Clean exit confirmed:** that `FAILURE` + `coverage_all_complete` combination is exactly what `CoverageEnded` → `IsCoverageComplete` → `CoverageCompleteDock` consumes, i.e. **mow area N → `MOWING_COMPLETE` → dock**, not `COVERAGE_FAILED_DOCKING`.
* **Cleared in two places, both needed** (justified in comments at each site):
  * `EndSession` — the real session boundary, alongside every other per-session set. The normal path.
  * the `~/high_level_control` handler on a plain `COMMAND_START` — a session can legitimately outlive a dock (low-battery auto-resume, emergency keep it alive so mowing resumes), so a stale clip could still be latched when the operator presses the ordinary Start button expecting the whole lawn. Safe: the GUI's per-area button calls `~/start_in_area`, which sets `current_command` itself and never reaches that handler, so this cannot cancel a targeted request. It also drops an unconsumed `target_area_index` so a request that was never picked up can't hijack a later plain start.
* **Persisted** in the coverage resume snapshot next to `current_command`: that restored command auto-re-enters `MowingSequence` after a container restart, so without persisting the clip a restarted targeted run silently widened into a whole-lawn run again. An absent line = not targeted (backward compatible).

## Tests

`test_get_next_unmowed_area.cpp` (real in-process `get_mowing_area` service, real BT ticks):

* `TargetedRunDoesNotRollOverToTheNextArea` — the dispatch after the target completes selects nothing and reports `coverage_all_complete` (**fails on `dev`**: it selects area 0)
* `TargetedRunReMowsAnAlreadyCompletedArea` — the existing completed/attempted erase is preserved
* `TargetedRunDoesNotReMowItsOwnCompletedArea` — the erase does not repeat on re-entry
* `PlainStartAfterATargetedRunIteratesAllAreas` — via the production `clearSingleAreaMode`
* `EndSessionClearsSingleAreaMode` — ticks the real `EndSession` node

`test_coverage_persistence.cpp`: `RoundTripsSingleAreaTarget`, `AbsentSingleAreaTargetMeansAllAreas`.

`status_nodes.cpp` + `status_snapshot.cpp` were added to the `test_get_next_unmowed_area` target so the session-end assertion exercises the real `EndSession` rather than a hand-rolled reset.

## Safety

No change to blade commands, motion, or firmware interaction. The behavioural change is strictly *narrowing*: a targeted run now stops where the operator asked instead of mowing an area they did not select.

## Test plan

- [ ] CI green on `feat/targeted-mow-single-area`
- [ ] Robot: GUI → "mow this area" on a non-zero area → confirm the log shows the targeted-run line on the follow-up dispatch and then `targeted area N finished — ending the run`, followed by `MOWING_COMPLETE` + dock (not another area, not `COVERAGE_FAILED_DOCKING`)
- [ ] Robot: plain Start afterwards mows all areas from index 0
- [ ] Robot: targeted run interrupted by a container restart stays targeted

### DIFF
--- a/ros2/src/mowgli_behavior/CMakeLists.txt
+++ b/ros2/src/mowgli_behavior/CMakeLists.txt
@@ -226,10 +226,16 @@ if(BUILD_TESTING)
   # real in-process get_mowing_area service returning a nav-only area + a mowing
   # area and ticks GetNextUnmowedArea to confirm the nav-only zone is never
   # selected for coverage.
+  # Also covers targeted runs (~/start_in_area): "mow only area N" is session
+  # state, so the run must not roll over into another area when the BT re-enters
+  # MowingSequence after the target completes. status_nodes/status_snapshot are
+  # linked in for the real EndSession node (the session boundary that clears it).
   ament_add_gtest(test_get_next_unmowed_area
     test/test_get_next_unmowed_area.cpp
     src/coverage_nodes.cpp
     src/coverage_persistence.cpp
+    src/status_nodes.cpp
+    src/status_snapshot.cpp
   )
 
   target_include_directories(test_get_next_unmowed_area PUBLIC
--- a/ros2/src/mowgli_behavior/include/mowgli_behavior/bt_context.hpp
+++ b/ros2/src/mowgli_behavior/include/mowgli_behavior/bt_context.hpp
@@ -78,7 +78,8 @@ struct BTContext
   /// BT condition/action nodes.  Use std::lock_guard for RAII locking.
   ///
   /// Does NOT cover the coverage-tracking fields below (command state +
-  /// swath-completion model: target_area_index, attempted_areas,
+  /// swath-completion model: target_area_index, single_area_target,
+  /// attempted_areas,
   /// area_attempt_count, area_last_coverage, area_completed_swaths,
   /// area_swath_count, area_resume_pose_index, area_path_pose_count,
   /// area_plan_fingerprint, completed_areas, coverage_all_complete). Those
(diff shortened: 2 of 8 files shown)


fwelvering:
Unfortunattely the mower does not return to it's dock after a targeted mow but stays in "Mowing complete" on the field.
I think it should (try) to go back to it's docker after the mow is done.

PR #523, opened 2026-09-02
fix(hardware): harden runtime parameters and battery-state semantics

liltaket:
## Summary

- Harden startup-only hardware parameters. Timer rates and wheel track now remain read-only after declaration; runtime mutation is rejected because timers and kinematics are constructed only at startup.
- Timer cadence uses `std::chrono::nanoseconds`. Non-finite, non-positive, unrepresentable, and zero-duration timer rates are rejected. Finite, representable rates outside each task's preferred operating range are clamped at startup with a warning that states the parameter, requested value, and effective value. Defaults still yield 250 ms (heartbeat), 10 ms (publish), 500 ms (high-level), and 100 ms (dig monitor).
- `wheel_track` is finite-only and clamped to `[0.15, 0.60]` m, matching the STM32 firmware clamp. The effective value is used by both host odometry and the firmware configuration packet.
- `serial_rx_timeout_s = 0` deliberately disables its watchdog. Dig freshness/escape timeouts are explicitly finite and strictly positive; their descriptors do not claim a pseudo-operational `double::min()` floor.
- Publish real-firmware `BatteryState.percentage` as `NaN` while no verified state-of-charge source exists, and serialize that unknown value as JSON `null` in the session monitor. Voltage, current/charging state, packet layout, and status publishing are unchanged.

## Validation

- Added tests for representable timer boundaries/defaults, operational clamps, invalid rates, wheel-track clamp bounds, timeout contracts, battery unknown semantics, and actual ROS rejection of a runtime `publish_rate` update.
- Rebased onto current `dev` (`14e8ca26`).
- `git diff --check` and Python syntax checking pass locally. This Windows host lacks the ROS/clang-format toolchain and Docker's Linux engine, so the post-push GitHub Actions run is the authoritative build/test verification.

### DIFF
--- a/firmware/stm32/ros_usbnode/include/mowgli_protocol.h
+++ b/firmware/stm32/ros_usbnode/include/mowgli_protocol.h
@@ -312,7 +312,7 @@ typedef struct {
   float v_charge;                       /**< Charge input voltage [V] */
   float v_system;                       /**< Battery / system voltage [V] */
   float charging_current;               /**< Charging current [A] */
-  uint8_t batt_percentage;              /**< Battery state of charge [0-100] */
+  uint8_t batt_percentage;              /**< Reserved until a verified SoC provider exists; currently 0. */
   uint16_t crc; /**< CRC-16 CCITT over preceding bytes */
 } pkt_status_t;
 
--- a/firmware/stm32/ros_usbnode/src/ros/ros_custom/cpp_main.cpp
+++ b/firmware/stm32/ros_usbnode/src/ros/ros_custom/cpp_main.cpp
@@ -1337,7 +1337,10 @@ extern "C" void broadcast_handler() {
     status_pkt.v_charge = charge_voltage;
     status_pkt.v_system = battery_voltage;
     status_pkt.charging_current = current;
-    status_pkt.batt_percentage = 0; // TODO: compute from voltage curve
+    // No verified SoC source exists.  Keep the legacy byte for wire
+    // compatibility; the ROS bridge publishes percentage as unknown (NaN),
+    // not 0 %.  Do not infer SoC from voltage here.
+    status_pkt.batt_percentage = 0;
 
     pkt_reset_cause_t reset_pkt;
     reset_pkt.type = PKT_ID_RESET_CAUSE;
--- a/ros2/scripts/mow_session_monitor.py
+++ b/ros2/scripts/mow_session_monitor.py
@@ -549,7 +549,10 @@ class MowSessionMonitor(Node):
     def _battery_cb(self, msg: BatteryState) -> None:
         with self.state_lock:
             self.state.battery_voltage = float(msg.voltage)
-            self.state.battery_percentage = float(msg.percentage)
+            percentage = float(msg.percentage)
+            # ROS uses NaN for an unmeasured BatteryState percentage; JSON does
+            # not have a NaN value, so preserve that meaning as JSON null.
+            self.state.battery_percentage = percentage if math.isfinite(percentage) else None
 
     def _cmd_vel_nav_cb(self, msg: TwistStamped) -> None:
         with self.state_lock:
--- a/ros2/src/mowgli_hardware/CMakeLists.txt
+++ b/ros2/src/mowgli_hardware/CMakeLists.txt
@@ -233,6 +233,30 @@ if(BUILD_TESTING)
     PRIVATE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   )
+
+  # ---- test_hardware_bridge_parameters ----
+  # Includes the node implementation to exercise ROS's actual read-only
+  # parameter contract while a missing serial device keeps the test hardware-free.
+  ament_add_gtest(test_hardware_bridge_parameters
+    test/test_hardware_bridge_parameters.cpp
+    src/odometry_publisher.cpp
+  )
+  target_link_libraries(test_hardware_bridge_parameters
+    mowgli_hardware_core
+  )
+  target_include_directories(test_hardware_bridge_parameters
+    PRIVATE
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  )
+  ament_target_dependencies(test_hardware_bridge_parameters
+    rclcpp
+    std_msgs
+    std_srvs
+    sensor_msgs
+    geometry_msgs
+    nav_msgs
+    mowgli_interfaces
+  )
 endif()
 
 # ---------------------------------------------------------------------------
--- a/ros2/src/mowgli_hardware/CMakeLists.txt
+++ b/ros2/src/mowgli_hardware/CMakeLists.txt
@@ -215,6 +215,24 @@ if(BUILD_TESTING)
     PRIVATE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   )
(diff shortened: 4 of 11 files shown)


cedbossneo (changes requested):
Thanks for this — the battery half is correct and the zero-period-timer fix is a genuine bug catch. I'm requesting changes on the parameter half, where one issue is serious enough to block.

## Blocking: an out-of-range parameter now kills `hardware_bridge` permanently

`validate_startup_parameters()` (`hardware_bridge_node.cpp:753-772`) throws `std::invalid_argument`, and the new descriptors (`:364-395`, `:719-740`) throw `InvalidParameterValueException` even earlier. `main()` has no `try` — it is `rclcpp::spin(std::make_shared<HardwareBridgeNode>())` — so the exception escapes `main` and `std::terminate`s. Under the compose restart policy that is an **infinite crash loop of the node that owns the serial link, `~/emergency_stop`, odometry, blade commands and all telemetry**.

Concrete: someone sets `publish_rate: 5.0` in `hardware_bridge.yaml` to shave CPU. In range today (a 200 ms tick — degraded but running). After this PR the bridge never starts, and recovery requires SSH. The failure is fail-safe in the strict sense — heartbeat timeout makes firmware latch emergency — so it is not a *safety* regression, but it is squarely the "strand the operator" case.

## Blocking: the `wheel_track` bound does the opposite of what its own comment argues

`timer_period.hpp:19-24` says the range *"mirrors the firmware's `PKT_ID_SET_KINEMATICS` wheel-base clamp"* so that host odometry and firmware IK cannot use different geometry. But the firmware is `pid_constrain(pkt->wheel_base, 0.15f, 0.60f)` (`cpp_main.cpp:561`) — it **clamps and keeps running**.

So with `wheel_track: 0.14`: today firmware uses 0.15 and the host 0.14, a 7 % odometry divergence. With this PR: a dead bridge. Throwing achieves neither the stated goal nor a live robot. **Clamp to the same range and log loudly** — that actually eliminates the divergence *and* keeps the machine alive, which is what the comment says you want.

I'd apply the same reasoning to the four timer rates: clamp-and-warn for *operational* bounds, and reserve the throw for values that are genuinely unusable (non-finite, non-positive, or a period that rounds to zero).

## Non-blocking

3. **Every operational bound is duplicated as a magic literal.** `1.0/50.0`, `10.0/500.0`, `1.0/20.0` appear once in the descriptors (`:364-365, 378, 734`) and again in `require_operational_timer_rate` (`:762-765`). Change one and not the other and they silently disagree. Named constants shared by both call sites.

4. **`std::numeric_limits<double>::min()` as a "strictly positive" floor is misleading.** At `:719-740` the dig timeouts get a lower bound of ~2.2e-308. It rejects `0.0` as intended, but `dig_pose_timeout_s: 1e-300` passes both the descriptor and `require_finite_positive` and permanently stands the detector down — the exact failure the comment says it prevents. Use a real operational floor, or `denorm_min()` with a comment.

5. **The headline read-only claim is untested.** Nothing asserts that a runtime `set_parameter` on `publish_rate` is rejected. `test_timer_period.cpp` on the other hand is genuinely good — boundaries, `nextafter` either side of the minimum, exact 1 ns, NaN and ±Inf.

6. **Battery `NaN` reaches the session monitor as a bare `NaN` token.** `mow_session_monitor.py:552` does `float(msg.percentage)` unguarded and `:864` serialises with `json.dumps`; `default=` is not consulted for floats, so the JSONL timeline emits `NaN`, which `JSON.parse` rejects. That file is a documented workflow (`docs/claude/session-monitoring.md`), so a `None` coercion in `_battery_cb` is worth adding.

7. `mowgli_simulation/src/fake_hardware_bridge_node.cpp:235` still publishes `percentage = 1.0f`, so sim and real now disagree about the field's meaning. Harmless today, nothing consumes it.

## What I verified in your favour

- **The NaN percentage is spec-correct and safe.** `sensor_msgs/BatteryState` documents NaN for an unmeasured percentage and firmware hardwires `0` (`cpp_main.cpp:1318`). I checked every consumer: the only in-repo subscriber is the session monitor; the BT's `battery_percent` is voltage-derived (`behavior_tree_node.cpp:220-228`) and `mowgli_leds` reads `HighLevelStatus`. I also disassembled `libsimple_charging_dock.so` on the robot — it references `charging_threshold` and `use_battery_status` and compares **current**, never percentage, consistent with CLAUDE.md invariant 12. Nothing safety-relevant sees it. Anything that previously did `percentage < threshold` was already firing permanently against a hardwired `0`.
- Invariant 12 intact: `msg.current` and `power_supply_status` unchanged at `:1484-1490`.
- The zero-period hazard was real: `std::chrono::milliseconds(static_cast<int>(1000.0 / rate))` truncates to `0 ms` above 1000 Hz. The nanosecond conversion is the right fix.
- No behaviour change at any shipped config — 4/100/2/10 Hz and the installed `publish_rate: 50.0` all give byte-identical periods.

## Logistics

This conflicts with #529, #524 and #536 — all four append an `ament_add_gtest` block at the same anchor in `mowgli_hardware/CMakeLists.txt`. It also branches from `c1818174`, before #537 landed. A rebase will be needed, and given the startup-failure modes above I'd want fresh CI on the rebased version specifically.

cedbossneo (approved):
Re-reviewed after the four follow-up commits. My blocking concern is resolved and I verified the battery change does not reach any safety path.

**On the crash-loop concern (was blocking).** `18761845` splits the handling correctly: values that are merely out of the preferred range are now clamped with a startup warning, and only genuinely unrepresentable input throws. I checked what that leaves reachable:

- `wheel_track` is the only one of these that is operator-facing (present in both the `mowgli_bringup` template and `gui/asserts/mower_config.schema.json`). It throws only on non-finite input, which a GUI number field cannot produce; every finite value clamps into `[0.15, 0.60]`. A `wheel_track: 0` typo now clamps instead of killing the node.
- `heartbeat_rate`, `publish_rate`, `high_level_rate` and `serial_rx_timeout_s` have no template entry and no schema entry, so they are not reachable from the GUI at all. A developer forcing an unrepresentable timer period getting a loud failure is the right outcome.

That is a good split — I would not want a NaN `wheel_track` silently clamped on a machine with blades.

**On the battery semantics.** Publishing `BatteryState.percentage` as NaN instead of a fabricated 0 % is correct per the message spec, and I traced every consumer before agreeing:

- The behaviour tree does **not** read this field. `behavior_tree_node.cpp` derives `battery_percent` from the filtered rail voltage via `batteryPercentFromVoltage()`, so `battery_critical_percent` / `battery_full_percent` hysteresis is untouched.
- `/battery_state` exists for `opennav_docking`. `simple_charging_dock` runs with `use_battery_status: true` and `charging_threshold: 0.3`, which is an ampere threshold read from the current field — the same field root `CLAUDE.md` invariant 12 pins as `abs(charging_current)`. Percentage is unread there.
- Nothing in `gui/` subscribes to `BatteryState`, so there is no Go `encoding/json` NaN marshalling hazard. `mow_session_monitor.py` is handled in `739bcd96`.

So the 0 % → NaN change is confined to a field no consumer interprets. Worth stating explicitly in the PR body, because "battery reads unknown instead of empty" sounds alarming until you trace it.

**Firmware.** `cpp_main.cpp` and `mowgli_protocol.h` are comment-only — `status_pkt.batt_percentage = 0;` is byte-identical before and after. No wire-format or behavioural change, which is what I wanted to see on a safety-authority file.

CI is green across all ten checks. Approving.

PR #524, opened 2026-09-02
fix(safety): reject non-finite velocity commands end-to-end

liltaket:
## Summary

Reject non-finite `cmd_vel` values (`NaN`, `+Inf`, `-Inf`) at both safety boundaries without changing finite-command shaping, PID gains, velocity limits, teleop cadence, twist_mux, or blade/e-stop behaviour.

## Root cause

`hardware_bridge_node` previously forwarded `linear.x` and `angular.z` to the float32 `LlCmdVel` packet without a finite check. STM32 refreshed `last_cmd_vel_tick` before validating either field and then relied on ordinary `>`/`<` clamps. `NaN` compares false to both bounds, so those clamps cannot catch it; a value could contaminate `cmd_wz` and wheel targets while an invalid packet refreshed the watchdog.

## Safety contract (before / after)

Before, a non-finite host message or valid-CRC raw wire packet could enter the motion path. After:

- The ROS2 callback rejects non-finite commands before mode inference, shaping, or host-state bookkeeping. It throttles diagnostics and discards the packet; it does **not** send a synthetic zero, because that would refresh the firmware's valid-command watchdog.
- The final `LlCmdVel` construction boundary independently requires finite `double` inputs within the finite float32 range **before** narrowing. Thus the `double → float` cast only occurs after representability is established; this avoids relying on an out-of-range floating conversion to produce `Inf`.
- Firmware independently validates both wire fields before touching `last_cmd_vel_tick`. Invalid packets zero yaw and both wheel targets, preserve the last valid timestamp, and return before inverse kinematics. A following finite command is accepted normally.

This is defense in depth: ROS2 protects the normal publisher path and host state; firmware remains the final authority for malformed/direct wire traffic and future host regressions. The wire layout and protocol version are unchanged.

## Motion-path review

All production `LlCmdVel` construction sites were searched; the bridge packet boundary is the sole constructor. Firmware's receive handler is separately guarded. Invalid direct-wire input clears targets before the next motor-control snapshot; the established wheel PI/yaw zero-command behavior performs bounded closed-loop deceleration/reset handling, with no stale command retained. No `-ffast-math` or `-ffinite-math-only` production flag is present.

## Tests

- ROS2 gtest covers finite commands, `+0`/`-0`, zero, NaN, `+Inf`, and `-Inf` for each axis; both axes non-finite; `DBL_MAX`/`-DBL_MAX`; exact `+/-FLT_MAX`; and the immediately out-of-range `std::nextafter` values. Rejected values are never cast to float.
- Firmware-native Unity test covers all NaN/+Inf/-Inf axis combinations, immediate target clearing, no watchdog timestamp refresh, zero, preservation of existing wheel targets on the helper's valid path, and valid-after-invalid behavior.
- Firmware CI runs the native test on Linux plus Yardforce500 and Yardforce500B builds.
- The complete ROS2 CI runs formatting, cppcheck, config drift, workspace build, `colcon test`, and the external GNSS launch contract.

No physical mower is needed to verify this logic. Physical actuation and watchdog timing remain separate hardware-level validation before field deployment.

## Remaining risk

Host-discarded invalid input relies on the established 200 ms firmware watchdog; direct-wire invalid input also zeros targets immediately. This PR remains Draft until the full CI result for the final commit is green.

### DIFF
--- a/.github/workflows/firmware-ci.yml
+++ b/.github/workflows/firmware-ci.yml
@@ -61,6 +61,23 @@ jobs:
         run: pio run -e ${{ matrix.environment }}
         working-directory: firmware/stm32/ros_usbnode
 
+  test:
+    name: Firmware safety unit tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.11'
+
+      - name: Install PlatformIO
+        run: pip install platformio
+
+      - name: Run native firmware unit tests
+        run: pio test -e native
+        working-directory: firmware/stm32/ros_usbnode
+
   # ─── Publish prebuilt binaries on a version tag ────────────────────────────
   # Phase 1 of the precompiled-firmware plan. On a `v*.*.*` tag, build every
   # default permutation and attach the stamped, checksummed .bin/.elf plus a
--- a/firmware/stm32/ros_usbnode/platformio.ini
+++ b/firmware/stm32/ros_usbnode/platformio.ini
@@ -16,6 +16,12 @@ build_src_filter =
 	-<ros/ros_lib/**>
 	-<ros/ros_lib>
 
+[env:native]
+platform = native
+framework =
+test_build_src = false
+build_flags = -Isrc/ros/ros_custom
+
 [env:Yardforce500]
 board = genericSTM32F103VC
 platform_packages =
--- a/firmware/stm32/ros_usbnode/src/ros/ros_custom/cmd_vel_safety.hpp
+++ b/firmware/stm32/ros_usbnode/src/ros/ros_custom/cmd_vel_safety.hpp
@@ -1,0 +1,32 @@
+#pragma once
+
+#include <cmath>
+#include <cstdint>
+
+namespace mowgli_cmd_vel
+{
+
+struct SafetyState
+{
+  float cmd_wz;
+  float left_target_mps;
+  float right_target_mps;
+  uint32_t last_valid_tick;
+};
+
+/// Accept only finite wire values; invalid input clears targets but preserves
+/// the last-valid timestamp so it cannot refresh the motion watchdog.
+inline bool apply_safety(float vx, float wz, uint32_t tick, SafetyState &state)
+{
+  if (!std::isfinite(vx) || !std::isfinite(wz)) {
+    state.cmd_wz = 0.0f;
+    state.left_target_mps = 0.0f;
+    state.right_target_mps = 0.0f;
+    return false;
+  }
+  state.cmd_wz = wz;
+  state.last_valid_tick = tick;
+  return true;
+}
+
+}  // namespace mowgli_cmd_vel
--- a/firmware/stm32/ros_usbnode/src/ros/ros_custom/cpp_main.cpp
+++ b/firmware/stm32/ros_usbnode/src/ros/ros_custom/cpp_main.cpp
@@ -40,6 +40,7 @@
 // COBS protocol (replaces rosserial)
 #include "mowgli_comms.h"
 #include "mowgli_protocol.h"
+#include "cmd_vel_safety.hpp"
 
 // Math
 #include <cmath>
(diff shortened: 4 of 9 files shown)


cedbossneo (approved):
Approving. The vulnerability is real and the firmware fix is correctly ordered. Two things worth deciding before or shortly after merge, neither blocking.

## Verified

The hole is genuine. On `dev`, `on_cmd_vel` sets `last_cmd_vel_tick = HAL_GetTick()` **before** any validation (`cpp_main.cpp:406`), and the only guards are `>`/`<` clamps (`:429-436`) — which NaN passes. So a NaN `angular.z` produced NaN wheel targets *and* refreshed the very watchdog that would otherwise have stopped the robot. That is the worst shape a bad input can take.

The refactor preserves ordering. Here (`:407-430`): validate → `last_cmd_vel_tick = …` → `if (IDLE) return;`, versus `dev`'s tick → IDLE-return. **The IDLE path still refreshes the tick**, so no behaviour change there. Invalid packets zero `cmd_wz` / `left_target_mps` / `right_target_mps` and preserve the old tick, so the 200 ms watchdog in `motors_handler` (`:846-849`) fires. I checked that watchdog only sets `hard_stop` and does **not** latch an emergency, so a transient NaN cannot strand anyone.

Concurrency is fine: those three globals are `volatile` (`:79-80, 242`), `on_cmd_vel` runs in USB RX interrupt context, and `motors_handler` snapshots under `__disable_irq()` (`:815-822`), so the ISR's writes cannot tear against the snapshot.

`send_cmd_vel_packet` (`hardware_bridge_node.cpp:2997`) is indeed the sole `LlCmdVel` construction site — I grepped to confirm. Checking float32 representability *before* narrowing, rather than casting and inspecting the result, is the right call.

Both changes are strictly motion-reducing, so the host still only ever removes motion and firmware stays the final authority.

## Worth deciding

1. **The ROS-side discard leaves the previous non-zero target live for up to 200 ms.** `hardware_bridge_node.cpp:2633-2645` returns without sending, so firmware never sees the invalid packet and its new zeroing path never runs — it coasts on the last valid target until the watchdog trips. Your description rejects sending a synthetic zero "because that would refresh the firmware's valid-command watchdog." I think that reasoning inverts for this case: a zero command *is* a stop, so refreshing the watchdog with one is harmless, and it stops the robot in ~0 ms instead of ≤200 ms — roughly 8 cm at 0.4 m/s. Either way, worth making the decision explicit in a comment rather than leaving it implied.

2. **`apply_safety`'s contract is fragile and half-dead.** In `cmd_vel_safety.hpp` the valid path sets `state.cmd_wz = wz`, but the caller only writes `state` back on the **false** branch (`cpp_main.cpp:417-421`), so that assignment is dead. Meanwhile the valid path never touches `left/right_target_mps` because the IK sets them later. The result is a helper whose struct fields are meaningful in only one of two branches, with a write-back rule the caller must remember. Someone who later "tidies" this by writing back unconditionally would reset the wheel targets to their pre-IK values on every command. Either split it into `is_valid_cmd_vel(vx, wz)` plus an explicit `clear_motion_targets()`, or make the write-back total.

3. **No test covers the wiring, and the wiring owns the headline property.** The Unity test asserts "invalid input does not refresh `last_valid_tick`" — but `apply_safety` does not own `last_cmd_vel_tick`; the caller does, at `:424`. A regression that put `last_cmd_vel_tick = HAL_GetTick()` back at the top of the handler would pass every test here. Same on the ROS side: the two pure predicates in `test_cmd_vel_validation.cpp` are non-vacuous (a revert fails them), but nothing proves `on_cmd_vel` discards or that `send_cmd_vel_packet` refuses.

I checked and cleared two CI risks the `[env:native]` addition raises: `[platformio] default_envs = Yardforce500` so a bare `pio run` does not build it, and the GUI flashes with `platformio run -e <env> -t upload` (`firmware.go:169`), so the flash path is unaffected.

## Logistics

Conflicts with #529, #523 and #536 at the same `ament_add_gtest` anchor in `mowgli_hardware/CMakeLists.txt`, and branches from `f21729e9`, before #537. Please rebase; I'll merge after #536 and #529.

PR #529, opened 2026-09-03
fix(safety): make emergency stop dominate release

liltaket:
## Safety fix

A malformed or buggy heartbeat may carry both `emergency_requested` and `emergency_release_requested`. The prior firmware processed them as independent sequential `if` statements: it asserted STOP, then could immediately process RELEASE and clear the new latch when no physical sensor was active.

This PR makes the heartbeat decision explicit: **STOP ALWAYS DOMINATES RELEASE.** Both fields are wire flags (`non-zero` is true); therefore a malformed non-zero value cannot create another command meaning.

| emergency_requested | emergency_release_requested | result |
| --- | --- | --- |
| 0 | 0 | unchanged, except the existing pure watchdog-latch recovery |
| 1 | 0 | emergency asserted |
| 0 | 1 | release only when no physical emergency is active |
| 1 | 1 | emergency asserted — STOP wins |

### Watchdog and physical-safety behavior

The existing watchdog-only recovery remains intact: a normal heartbeat clears a pure communications-loss latch only when no physical sensor is active. If a physical input appears while that latch is pending, firmware keeps the emergency state asserted and drops its watchdog-only provenance. An explicit STOP is evaluated first, so neither watchdog recovery nor a simultaneous RELEASE can clear it.

`any_physical_emergency()` is unchanged and still covers yellow/white stop buttons, blue/red wheel lift, mechanical tilt, and low-Z accelerometer input. A host heartbeat cannot release any currently active physical emergency.

The host deliberately has independent STOP and one-shot RELEASE state: startup/reconnect watchdog recovery and lift recovery can request RELEASE while an emergency becomes active. This PR makes that combination deterministic in firmware rather than depending on host timing.

### Tests and CI

- `test_heartbeat_emergency_policy`: 13 GTests, including the complete 16-case truth table, STOP+RELEASE with/without physical input, release after a prior STOP, normal watchdog recovery, physical emergency after a watchdog latch, and watchdog-latch + RELEASE precedence.
- The policy helper uses the exact firmware C++11 toolchain and builds for both `Yardforce500` and `Yardforce500B`.
- ROS2 CI now treats exactly `firmware/stm32/ros_usbnode/include/heartbeat_emergency_policy.hpp` as ROS2-relevant, both on `push.paths` and PR change detection. This runs the policy GTest for header-only changes without making unrelated firmware changes build ROS2.
- Verified in GitHub Actions: Firmware builds/board-defaults guard, ROS2 change detection, format, cppcheck, full ROS2 build/test. The full ROS2 log shows `test_heartbeat_emergency_policy` built and all 13 tests passed.
- Local checks: both firmware targets, board-defaults guard, protocol-version guard, clang-format for the new test, and `git diff --check`.

No physical mower test is required for correctness of this isolated deterministic precedence policy. This does not claim to replace hardware verification of the wider emergency system.

### DIFF
--- a/.github/workflows/ros2-ci.yml
+++ b/.github/workflows/ros2-ci.yml
@@ -7,6 +7,7 @@ on:
       - 'ros2/**'
       - 'tools/motor/**'
       - 'install/config/mowgli/**'
+      - 'firmware/stm32/ros_usbnode/include/heartbeat_emergency_policy.hpp'
       - '.github/workflows/ros2-ci.yml'
   # NOTE: deliberately NO `paths:` filter here, and `dev` is included alongside
   # `main`. Two bugs used to hide in this block:
@@ -74,6 +75,7 @@ jobs:
           fi
           CHANGED=$(git diff --name-only "$BASE" HEAD -- \
             'ros2/**' 'tools/motor/**' 'install/config/mowgli/**' \
+            'firmware/stm32/ros_usbnode/include/heartbeat_emergency_policy.hpp' \
             '.github/workflows/ros2-ci.yml')
           if [ -n "$CHANGED" ]; then
             echo "ROS2-relevant changes:"; echo "$CHANGED"
--- a/firmware/stm32/ros_usbnode/include/heartbeat_emergency_policy.hpp
+++ b/firmware/stm32/ros_usbnode/include/heartbeat_emergency_policy.hpp
@@ -1,0 +1,45 @@
+// Copyright 2026 Mowgli Project
+// SPDX-License-Identifier: GPL-3.0-or-later
+//
+// Pure heartbeat emergency policy.  Keep this independent of hardware so the
+// host/firmware boundary's stop-over-release invariant is unit-testable.
+
+#ifndef HEARTBEAT_EMERGENCY_POLICY_HPP
+#define HEARTBEAT_EMERGENCY_POLICY_HPP
+
+enum class HeartbeatEmergencyAction {
+  Unchanged,
+  Assert,
+  Release,
+  AutoClearWatchdog,
+};
+
+struct HeartbeatEmergencyDecision {
+  HeartbeatEmergencyAction action;
+  bool clear_heartbeat_only_latch;
+};
+
+// STOP ALWAYS DOMINATES RELEASE. A watchdog-only latch may still auto-clear on
+// a normal heartbeat, but never when that heartbeat explicitly requests STOP.
+constexpr HeartbeatEmergencyDecision decide_heartbeat_emergency(
+    const bool emergency_requested, const bool emergency_release_requested,
+    const bool physical_emergency, const bool heartbeat_only_latch) {
+  return emergency_requested
+             ? HeartbeatEmergencyDecision{HeartbeatEmergencyAction::Assert,
+                                          true}
+         : heartbeat_only_latch
+             ? (physical_emergency
+                    ? HeartbeatEmergencyDecision{HeartbeatEmergencyAction::
+                                                     Unchanged,
+                                                 true}
+                    : HeartbeatEmergencyDecision{HeartbeatEmergencyAction::
+                                                     AutoClearWatchdog,
+                                                 true})
+         : emergency_release_requested && !physical_emergency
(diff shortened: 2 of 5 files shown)


cedbossneo (approved):
Approving. The hazard is real, the refactor is faithful, and firmware is the right place for the enforcement. Follow-ups below, none blocking.

## Verified

I rebuilt the old/new truth tables by hand from `cpp_main.cpp:363-395` on `dev` against `heartbeat_emergency_policy.hpp` + `cpp_main.cpp:366-402` here. **All 16 rows match**, except the intended `stop && release → Assert` and one cosmetic difference: `watchdog_latch && release && !physical` now logs "comms-loss emergency auto-cleared" where it previously logged nothing. Both call `Emergency_SetState(0)`, so no functional divergence.

The glue at `:373-376` is sound — `physical_emergency` is only consumed in branches already guarded by `watchdog_latch_active` or `release_requested`, so short-circuiting to `false` outside them cannot change a decision.

**I also found the concrete host-side sequence** your description only gestures at, which makes the case stronger than written. `emergency_active_` and `emergency_release_pending_` are not mutually exclusive (`hardware_bridge_node.cpp:2158-2159`), and `startup_release_count_{5}` (`:3184`) forces `emergency_release_pending_ = true` on the first five heartbeats — about 1.25 s at 4 Hz. An operator or BT calling `~/emergency_stop(emergency=1)` inside that window sets `emergency_active_` (`:2993`) while the burst is still running, so **one heartbeat carries both flags**. On `dev`'s firmware that asserts and then immediately releases, because no physical sensor is active at boot: the e-stop is silently swallowed. Narrow, but real, and this closes it.

I could not construct a stranding case. `ResetEmergency` (`utility_nodes.cpp:237-262`) and the GUI (`mowglinext.go:574`) both go through `~/emergency_stop`, and `on_emergency_stop(0)` clears `emergency_active_` *and* sets the release in the same call (`:2997-2999`), so there is always an exit from STOP dominance. Invariant 9's dock auto-reset is unaffected.

## Follow-ups

1. **The dominance is completely silent.** `:400-402` logs only the `!emergency_requested && physical` rejection. When STOP dominates a concurrent release nothing is logged on either side, and the host has already consumed its one-shot flag (`:2162`). "My reset didn't take" would be undiagnosable in the field. A one-line `debug_printf` in the `Assert` branch when `emergency_release_requested` is also set costs nothing.

2. **The root cause is host-side and left in place.** Firmware now masks a self-contradictory packet, but `hardware_bridge_node` still sends one. `emergency_active_` can never legitimately be true during the startup burst — it initialises `false` and only the service sets it — so refusing to set `emergency_release_requested` while `emergency_active_` is strictly correct and fixes it on the side that has the intent.

3. **The tests pin the policy but not the part that could go wrong.** The 13 gtests cover `decide_heartbeat_emergency` only, and they are not vacuous — a revert fails 8 truth-table rows plus both `StopDominates*` cases. But the *new* expression at `:373-376`, the one with real inversion risk, has no coverage, and nothing tests that the host can even produce `stop && release`. The scenario motivating the PR is untested end to end.

4. **Layering.** `mowgli_hardware/CMakeLists.txt` now adds `../../../firmware/stm32/ros_usbnode/include`. The production image is safe — `ros2/Dockerfile:407` builds `-DBUILD_TESTING=OFF` and never `COPY`s `firmware/` — but any `BUILD_TESTING=ON` build of `ros2/` alone will fail to compile. A shared header location would be cleaner.

5. A four-level nested ternary is a hard shape to read for a safety-precedence policy. Nothing needs `constexpr` here; dropping it allows an `if`/`else` chain that reads like your own truth table.

In your favour, unprompted: `any_physical_emergency()` reaches `I2C_TestZLowINT()` (`i2c.c:190`), a latch-clearing read of the LIS3DH `INT1_SRC` from USB RX interrupt context. This PR *reduces* how many of those happen per heartbeat. Neutral to better.

## Logistics

Conflicts with #524, #523 and #536 — all four append an `ament_add_gtest` at the same anchor in `mowgli_hardware/CMakeLists.txt`. Branches from `f21729e9`, before #537. Please rebase; I'll merge after #536.

Issue #533, opened 2026-09-04
A frozen GNSS receiver is indistinguishable from a healthy one: /gps/fix and /gps/status are republished on a timer, so every freshness check passes


cedbossneo:
## The mechanism, traced end to end

The upstream `receiver_node` decouples acquisition from publication (UGA-003). `PublishNow()` runs on `publication_timer_` at `publish_rate_hz` — **5 Hz on this robot** — and publishes a snapshot of the *current runtime state*:

```cpp
void PublishNow()
{
  auto snapshot = BuildSnapshot();
  last_status_message_ = std::move(snapshot.status);
  if (CanPublishFixMessage(snapshot.runtime_state))
  {
    last_fix_message_ = ToNavSatFixMessage(snapshot.runtime_state);
    fix_publisher_->publish(*last_fix_message_);
  }
  ...
  status_publisher_->publish(*last_status_message_);
}
```

Nothing here requires a *new* observation. If the receiver stops producing fixes — USB wedged, antenna disconnected, firmware hang — the node keeps emitting, at 5 Hz:

- `/gps/fix` with the **last known latitude/longitude**, and
- `/gps/status` with `rtk_mode = 3` and `horizontal_accuracy_m = 0.014` **frozen at their last values**.

## What that breaks downstream

| consumer | what it does | behaviour with a frozen receiver |
|---|---|---|
| `navsat_to_absolute_pose_node` | `/gps/fix` → `/gps/pose_cov` | keeps emitting a **frozen absolute position** at 5 Hz; no staleness guard exists in this node |
| `fusion_graph_node` | GNSS unary factor on `/gps/pose_cov` | anchors the map estimate to a position the robot has driven away from; no GPS-age guard exists |
| `hardware_bridge` → `DigTrustSigma` | `gnss_fresh && rtk_fixed && accuracy_valid` | **`gnss_fresh` measures message arrival, not observation novelty** (`have_dig_gnss_ && (tick_now - dig_gnss_time_) < dig_gnss_timeout_s`), so it stays true and the dig detector fully trusts a dead GPS |
| BT `LocalizationGuard` | GNSS solution quality from `/gps/status` | sees a healthy RTK-Fixed solution |
| GUI | `GnssLiveDiagnosticsCard` | shows RTK Fixed, 14 mm |

[...]

PR #534, opened 2026-09-04
Fix/gnss downstream

Pepeuch:
## Summary

Update and validate the MowgliNext Universal GNSS downstream integration against the current validated `universal-gnss/main` baseline.

This PR aligns MowgliNext with the latest Universal GNSS contracts, updates the vendored submodule to the validated `main` commit, and preserves the GNSS provenance/freshness semantics required by localization, fusion, hardware, and behavior consumers.

Universal GNSS target:

`f974b565100b4f2aa3d522cd9a292f30f905e8bc`

Previous MowgliNext pin:

`5281472116669972ae12b9d1997d66b064671cf5`

## Changes

- Update `ros2/src/external/universal-gnss` to validated Universal GNSS `main` commit `f974b56`.
- Change the Universal GNSS submodule tracking branch from `review` to `main`.
- Preserve receipt-time provenance between Universal GNSS and MowgliNext GNSS consumers.
- Preserve `position_observation_sequence` so cached publications are not interpreted as fresh receiver observations.
- Preserve the distinction between heading and course-over-ground.
- Preserve GNSS capability/value flags used by downstream consumers.
- Preserve RTK/correction-health semantics used by localization and diagnostics.
- Validate GNSS status/NavSat pairing and freshness behavior through the downstream sidecar.
- Validate fusion graph GNSS freshness integration.
- Validate hardware/localization consumers against the updated GNSS contract.
- Validate GNSS launch/configuration paths for receiver family, device, baud, and NTRIP configuration.
- Confirm no obsolete `frame_writer` integration is present.
- Confirm the GPS image build consumes Universal GNSS from the vendored submodule, so the image source and parent gitlink resolve to the same UG revision.
- Record downstream validation evidence in the local integration checkpoint.

No unrelated Universal GNSS redesign is included in this PR.

## Component

- [x] ROS2 Stack (`ros2/`)
- [ ] GUI (`gui/`)
- [ ] Docker (`docker/`)
- [x] Sensors (`sensors/`)
- [ ] Firmware (`firmware/`)
- [ ] Documentation (`docs/`)
- [ ] CI/CD (`.github/`)

## Testing

- [x] Built successfully
- [x] Tested on hardware
- [ ] Tested in simulation
- [x] Unit tests pass
- [x] Manual testing

Validation performed against Universal GNSS `f974b56`:

- Focused ROS2 build passed for:
  - Universal GNSS
  - localization
  - fusion graph
  - hardware bridge
  - behavior
  - bringup
- GNSS adapter tests passed.
- GNSS receiver/NTRIP integration tests passed.
- Fusion GNSS freshness tests passed.
- Hardware/localization GNSS consumer tests passed.
- Behavior GNSS-related tests passed.
- GNSS launch/configuration tests passed.
- NavSat integration tests passed.
- Downstream receipt-time, observation-sequence, heading/COG and correction-health contracts were inspected and validated.
- Universal GNSS submodule SHA matches the source consumed by `sensors/gps/Dockerfile`.
- `git diff --check` passes.

### Known unrelated limitations

The full workspace build is currently blocked by a pre-existing Fields2Cover dependency/version mismatch in the coverage-planning stack.

This does not affect the Universal GNSS downstream integration validated by this PR.
existing Fields2Cover dependency mismatch:

- installed: Fields2Cover 2.0
- required by current MowgliNext dependency chain: Fields2Cover 3.0

This is pre-existing and unrelated to the Universal GNSS update.

One Colcon TF-ownership test is sensitive to the working directory:
- it fails when launched from `/ros2_ws`
- it passes when executed from its intended `ros2` workspace root

Neither issue was introduced by this PR.

## Checklist

- [x] Documentation updated (if user-facing)
- [x] No unrelated changes


cedbossneo:
Thanks for this — the freshness core is genuinely good work, and it lands on a problem we filed independently two days ago (#533) after tracing exactly the failure mode you address: `receiver_node` republishes cached status and NavSatFix on its 5 Hz publication timer, so every freshness check we have downstream measures *message arrival* rather than *observation novelty*, and a frozen receiver keeps looking healthy to the dig detector, the localizer, the factor graph and the BT alike.

`gnss_observation_freshness.hpp` is better than what we had sketched for that issue. Separating ROS-clock receipt provenance from monotonic delivery liveness, and making source restart / cached publication / out-of-order / clock discontinuity distinct outcomes rather than one boolean, is the right shape — and the `sequence == 0` fallback keeps older sidecar images working. Wiring it through all four consumers rather than just one is what makes it useful.

## What we verified

Before deciding anything we re-ran the checks rather than reading the diff:

| surface | result |
|---|---|
| `gui/pkg/msgs/mowgli/types_generated.go` | regenerated correctly — re-running `generate_go_msgs.sh` produces no diff |
| `gui/web/src/types/ros.generated.ts` | regenerated correctly — same |
| `firmware/.../ros_lib/mower_msgs/GnssStatus.h` | **not regenerated** — `LC_ALL=C python3 firmware/scripts/sync_ros_lib.py` produces **146 lines** this PR does not contain |
| C++ bridge | **not weakened** — we initially misread the −201/+326 line counts as logic moving to Python; it is an extraction of inline helpers into `CorrectionDiagnosticTracker` in the same C++ package. Our mistake, and worth stating since the C++ bridge is what actually runs in production |

There is a `.github/workflows/msg-codegen-drift.yml` that exists precisely to catch the firmware gap. **It never ran on this PR** — only 4 of the usual ~16 checks executed (the GPS image builds and the labeler). That is a repository-side workflow-approval thing for fork PRs, not something you did wrong, but it does mean the "Built successfully / Unit tests pass" boxes have not been corroborated by our CI on this branch.

## Why we are splitting rather than merging as-is

Two reasons, neither about the GNSS work itself.

**Base.** This branches from `f21729e9`, before `dev` received the universal-gnss submodule re-pin (#531, merged as `64451ef1`). It conflicts on `.gitmodules`, `AGENTS.md` and `README.md`. Your pin `f974b56` does descend from the current `dev` pin `ab32f67`, so advancing is linear and safe — we would just rather do it as its own change, since `dev` already tracks `main`.

**Scope.** The component checkboxes tick ROS2 and Sensors, but the diff also carries roughly 4,000 lines outside that: `docs/status/mowgli_backlog.{json,svg}` (+3,265), `MOWGLINEXT_GNSS_AUDIT.md` (+1,217), `scripts/update_backlog_status.py` and its test (+722), a rewrite of `AGENTS.md` (+372/−15), `.agent/policies/*`, and ~10 `gui/` files including a new `gnss_authority.go` provider. Those are real decisions — adopting a contributor's agent framework and a backlog-tracking subsystem into this repo is a governance choice, not a GNSS fix — and they should be argued on their own merits rather than ride along with a change we want to merge quickly.

## The plan

Three PRs against `dev`, each independently reviewable, all crediting you with `Co-authored-by`:

1. **Observation freshness** — your header verbatim, `position_observation_sequence` only, the four consumer wirings, and all three codegen surfaces regenerated including the firmware one. Already in progress.
2. **Correction diagnostics** — `CorrectionDiagnosticTracker`, the `correction_*` message fields and the bridge wiring. Staying in C++; the Python bridge is kept updated as the reference implementation only.
3. **Localization additions** — `cog_observation_timing`, `navsat_status_association`, the monitor policy.

For the documentation: rather than dropping `MOWGLINEXT_GNSS_AUDIT.md` and the agent policies, we would like to mine them into the docs we already have — the GNSS findings into `docs/claude/ros2-specifics.md`, and whatever is useful from the policies into the existing `AGENTS.md` rather than replacing it. If you would rather drive that yourself, say so and we will leave it to you.

Nothing here is a rejection: the substance is being kept, credited, and merged. Happy to be corrected on any of the above, particularly if there is a reason the submodule bump needs to land together with the freshness work.

cedbossneo:
Merged in 3 differents prs

PR #544, opened 2026-09-05
fix: honor charging limits from advanced robot config

wjcloudy:
## Summary

Saving `max_charge_voltage` and `max_charge_current` in advanced robot configuration currently leaves the hardware bridge using 29.4 V / 1.2 A. Forward both settings from the merged robot config at startup, so an operator's 28.5 V / 1.8 A configuration reaches the existing firmware safety-limit packet after restarting the stack.

**Safety-critical: charging behavior.** Firmware still enforces its compiled board ceilings. The defaults, firmware, protocol and GUI schema are unchanged.

## Changes

- Add the two missing hardware-bridge launch parameters.
- Extend the existing launch-injection guards to verify custom values and fallback defaults on the correct node.

## Component

- [x] ROS2 Stack (`ros2/`)

## Testing

- [x] Unit tests pass: 51 tests across `test_launch_injection.py` and `test_robot_config_util.py`.
- Both new regression cases fail on unmodified upstream and pass with the fix.
- `git diff --check` passes.
- The same two-parameter forwarding was checked during stationary hardware migration: runtime parameters and bridge transmit logs reported 28.50 V / 1.80 A. No complete charge-cycle validation is claimed.
- Full ROS2 build/colcon testing remains for CI; the local host is Windows without a ROS2 build environment.

## Checklist

- [x] No unrelated changes
- [x] Physical behavior flagged as safety-critical above

### DIFF
--- a/ros2/src/mowgli_bringup/launch/mowgli.launch.py
+++ b/ros2/src/mowgli_bringup/launch/mowgli.launch.py
@@ -201,6 +201,9 @@ def generate_launch_description() -> LaunchDescription:
             # Allow command-line override of the serial port.
             {"serial_port": serial_port},
             {"use_sim_time": use_sim_time},
+            # Forward configured charge ceilings; firmware enforces its board limits.
+            {"max_charge_voltage": float(robot_params.get("max_charge_voltage", 29.4))},
+            {"max_charge_current": float(robot_params.get("max_charge_current", 1.2))},
             # Pass dock pose from robot config for dock position anchoring
             {"dock_pose_x": float(robot_params.get("dock_pose_x", 0.0))},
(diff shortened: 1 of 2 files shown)


cedbossneo (approved):
Confirmed the three things that make this safe: `hardware_bridge_node.cpp:391-392` already declares both parameters, `:2427-2428` already sends them to the firmware, and `mowgli_protocol.h:562` documents that the firmware clamps them to `(0, compiled ceiling]`. The host can therefore only lower the charge envelope, never raise it above the board limit. The launch file was simply not forwarding the configured value, so a user lowering the ceiling in the GUI was silently ignored. The AST-based test is a solid guard. Approving.

Issue #515, opened 2026-09-02
Docking: 4× 'made contact' at the dock pose but charge never starts; DockRobot contact-delta log compares against a (0,0,0) dock


cedbossneo:
## Observed (2026-09-02 ~15:15 UTC, low-battery return after a 73 min LiDAR mow, RTK Fixed, IMU alive)

`docking_server` (SimpleChargingDock) reported **"Made contact with dock, waiting for charge to start"** on all four attempts, each followed by "Timed out waiting for charge to start", then `error_code:906`. The BT's contact-delta log (PR #494) at each contact:

```
retry 0 at (6.207, 2.825) vs dock (0.000, 0.000, 0.00°) — along=+6.207 cross=+2.825 range=6.820
retry 1 at (6.287, 2.857) vs dock (0.000, 0.000, 0.00°)
retry 2 at (6.274, 2.829) vs dock (0.000, 0.000, 0.00°)
retry 3 at (6.279, 2.819) vs dock (0.000, 0.000, 0.00°)
```

Configured `dock_pose` is (6.273, 2.773, −48.8°). So the robot reached a point within 1–8 cm of the dock pose every time, the docking server believed it was in contact, and no charge current appeared. The operator then pushed the robot the last few cm by hand and charging started at (6.278, 2.753) — i.e. **~7 cm further along the approach** than where the contact was claimed.

## Two defects

1. **The #494 diagnostic is blind.** `BTContext::dock_x/dock_y/dock_yaw` are never written anywhere in `mowgli_behavior` (only read), so `ComputeDockContactDelta` always compares against (0, 0, 0) and the along/cross/range numbers are meaningless. `full_system.launch.py` forwards `dock_pose_x/y/yaw` to **map_server_node only**; `behavior_tree_node` declares a legacy string parameter `dock_pose` ("x;y;yaw", default "0.0;0.0;0.0") that the launch never sets and that only lands on the blackboard. `CalibrateHeadingFromUndock`'s `if (ctx->dock_yaw == 0.0)` guard (calibration_nodes.cpp:248) is bypassed for the same reason.
[...]

PR #546, opened 2026-09-05
feat(localization): apply GNSS observation identity to the localization consumers

cedbossneo:
**Split 3 of 3 of #534** — the GNSS observation-identity audit contributed by @Pepeuch. Landing the audit in tiers rather than merging it wholesale.

- **Split 1** (merged): `mowgli_interfaces/gnss_observation_freshness.hpp` + its `fusion_graph` / `mowgli_behavior` / `mowgli_hardware` consumers and the `position_observation_sequence` field.
- **Split 2** (in flight, `feat/gnss-correction-diagnostics`): `mowgli_gnss_bridge` correction diagnostics + the `correction_*` fields.
- **Split 3** (this PR): the `mowgli_localization` consumers and their launch injection. No message changes, so it is independent of split 2.

Three commits, one per subsystem, so they review and revert independently.

## 1. `cog_observation_timing` — COG sample timing from the receiver's physical rate

`cog_to_imu` dropped its baseline anchor whenever two `/gps/fix` samples were more than a hardcoded `max_sample_dt_s = 0.50` apart, measured on **callback arrival**. At `gnss_profile_rate_hz = 1` every sample exceeds it, so the anchor reset on arrival and COG yaw never accumulated a baseline; at 10 Hz the same constant spans five receiver periods, so a genuinely missed cycle folded silently into the baseline.

`navigation.launch.py` now reads the existing `gnss_profile_rate_hz` install-time key (the same value `start_gps.sh` uses to configure the receiver) and injects it as `physical_gnss_observation_rate_hz`; the deadline is derived as `1.5 / rate`. `max_sample_dt_s` stays an operator override but is validated against the physical period.

The decision moved onto **receipt provenance**, never arrival: a bounded receipt history separates a delayed duplicate (ignored) from an unseen stamp rewind (epoch reset). `StationaryLatchClock` gives the stationary-yaw latch two explicit clocks — ROS receipt for validity, steady elapsed for expiry — so a paused or rewound ROS clock can no longer keep a dead latch alive.

## 2. `navsat_status_association` — pair a fix with its own status

`navsat_to_absolute_pose_node` held one "latest status" and stamped whichever `GnssStatus` arrived most recently onto whatever `NavSatFix` it was projecting. The two topics are delivered independently, so the pairing was only accidentally correct. It fails one-directionally and silently: a stale status keeps asserting RTK-Fixed over a degraded fix, and `/gps/pose_cov` — which `map_server` averages into the dock pose under an RTK-quality gate — inherits authority the receiver never granted for that measurement.

Replaced with a bounded rendez-vous keyed on receipt provenance and nothing else (coordinates are never compared — a stationary robot legitimately repeats them). A pair is released only once its monotonic window closes, so one receipt seen with two different `position_observation_sequence` values fails closed instead of publishing the first arrival. There is deliberately **no** latest-status fallback: an unmatched fix still yields `/gps/absolute_pose` from its own `NavSatFix` status, but never typed RTK authority, so `/gps/pose_cov` stays silent. `test_navsat_status_universal` now pins exactly that.

### One deliberate deviation from #534 — with a reproduction

As contributed, the rule *"a repeated `position_observation_sequence` on a different receipt stamp is an identity violation"* fires on the receiver's **normal steady state**.

I checked the vendored driver rather than assuming. Universal GNSS publishes fix + status together from **one snapshot** on a timer (`publish_rate_hz`, 10 Hz default) — both stamps come from the same `GnssRuntimeState::timestamp_ns` through the same `ToRosTime()`, so they are bit-identical and the stamp is a valid join key. But that receipt clock advances on **any** accepted field merge — a satellite-count or CN0 update is enough (`gnss_runtime_aggregator.hpp`: `applied && update.timestamp_ns.has_value()`) — whereas `position_observation_sequence` advances only on an accepted **position** observation. Between two fixes the stamp therefore moves and the sequence does not.

Replaying that cadence (10 Hz publish / 5 Hz position) through the association as contributed:

```
ready(pose_cov)=0  fix_only(abs_pose)=0  ambiguous=100
```

**100 % of deliveries classified ambiguous** — both `/gps/absolute_pose` and `/gps/pose_cov` permanently silent, one unthrottled `RCLCPP_ERROR` per tick. That would have taken out `localization_monitor`, the GUI's GPS display and the dock-calibration gate on real hardware.

That case is a cached republication, which the design already names, so it is now classified `kCachedStatus`: the receipt is closed, nothing is emitted for it, and no other pending association is touched. The genuine violations still fail closed — a repeated sequence on an **earlier** stamp, and two sequences on one stamp. Same replay after the fix:

```
ready(pose_cov)=10  ambiguous=0   (10 physical observations over 20 publish ticks)
```

Three regression tests cover it, including the 10 Hz-publish / 5 Hz-position cadence end to end.

Worth noting as intended behaviour: `/gps/absolute_pose` now runs at the **position** rate (5 Hz) rather than the publish rate (10 Hz), and both pose topics gain the association window (`status_pairing_window_s`, 0.25 s) of latency. Stamps are preserved, and `fusion_graph` is unaffected (it subscribes to `/gps/fix` directly per Invariant 1).

## 3. `localization_monitor_policy` — receipt freshness vs delivery liveness

The monitor decided `DEAD_RECKONING` from the arrival stamp of the last `/gps/absolute_pose`. Delivery and physical observation are not the same signal here: the adapter republishes its last accepted fix on a timer with the receipt stamp unchanged, so a frozen receiver keeps the published mode looking healthy indefinitely.

`ObservationIsFresh()` now drives the mode while `DeliveryIsLive()` only shapes the log line, which distinguishes *"physical observations stale while cached message delivery remains active"* (frozen receiver) from *"observations and delivery both stale"* (dead link) — different operator actions. A clock-epoch break bypasses the `mode_debounce_sec_` hysteresis: it is not mode flicker, so the safe mode commits immediately. The mode ordering moved to a free function in `localization_monitor_policy.hpp` so it is unit-testable without a node.

## Verified

Built and tested in a `ros:kilted-ros-base` container (`colcon build --packages-up-to mowgli_localization`, Release):

| | result |
|---|---|
| `colcon test --packages-select mowgli_localization` | **171 tests, 0 failures, 0 errors**, 30 skipped (all skips are `cppcheck`, absent from the container) |
| new `test_cog_observation_timing` | 19 passed |
| new `test_localization_monitor_freshness` | 20 passed |
| `test_navsat_status_association` | 21 passed (18 from #534 + 3 new regressions) |
| `mowgli_bringup/test/test_launch_injection.py` | 7 passed |
| `mowgli_bringup/test/test_navsat_status_universal.launch.py` | 2 + shutdown, OK (live node, real DDS) |
| `ros2/scripts/format.sh` (clang-format **18.1.8**) | all 12 C++ files in this PR clean |

## Deliberately excluded

Nothing from `.agent/`, `AGENTS.md`, `README.md`, `MOWGLINEXT_GNSS_AUDIT.md`, `docs/status/`, `scripts/update_backlog_status.py`, `gui/`, `sensors/`, `.gitignore`, `.gitmodules`, `fusion_graph`, `mowgli_behavior`, `mowgli_hardware`, or `mowgli_interfaces`.

**The submodule gitlink `ros2/src/external/universal-gnss` is untouched** — `git diff origin/dev...HEAD -- ros2/src/external` is empty.

Two small notes:

- `navsat_to_absolute_pose_node.cpp`'s two TU-local helpers kept their `static` (the original dropped it, giving them external linkage in `namespace mowgli_localization`).
- `aa97b157 chore: auto-format C++ files` was generated by the repo's own pre-push hook, not by me. It reformats two **pre-existing** clang-format-18 violations in `mowgli_hardware/{src/serial_port.cpp, test/test_serial_port.cpp}` that are unrelated to this PR — any push from any branch triggers it. Whitespace only. Drop it if you'd rather it landed separately.

## Test plan

- [ ] Field: confirm `/gps/absolute_pose` and `/gps/pose_cov` keep publishing at the configured `gnss_profile_rate_hz` (the deviation above is reproduced in tests but has not run on hardware).
- [ ] Field: confirm `/localization/mode_id` still reports RTK-Fixed under a healthy receiver and drops to DEAD_RECKONING when the receiver is unplugged.
- [ ] Field: confirm COG yaw (`/imu/cog_heading`) still accumulates a baseline at the installed `gnss_profile_rate_hz`.
- [ ] Confirm one-click dock calibration still passes its `/gps/pose_cov` RTK gate.

### DIFF
--- a/ros2/src/mowgli_bringup/launch/navigation.launch.py
+++ b/ros2/src/mowgli_bringup/launch/navigation.launch.py
@@ -529,6 +529,11 @@ def generate_launch_description() -> LaunchDescription:
     declination_deg = 1.5
     min_horizontal_uT = 5.0
     mag_yaw_variance = 0.0027
+    # Physical receiver observation cadence, not the ROS publication cadence.
+    # The same gnss_profile_rate_hz value configures the receiver profile in
+    # start_gps.sh; pass it to cog_to_imu so its receipt-gap policy has one
+    # source of truth.
+    physical_gnss_observation_rate_hz = 5.0
     runtime_robot_config = "/ros2_ws/config/mowgli_robot.yaml"
     # Merged params: in-package template defaults with the installed sparse
     # config layered on top. Every rt_rp.get(key, <fallback>) below therefore
@@ -577,6 +582,12 @@ def generate_launch_description() -> LaunchDescription:
         declination_deg = float(rt_rp.get("declination_deg", declination_deg))
         min_horizontal_uT = float(rt_rp.get("min_horizontal_uT", min_horizontal_uT))
         mag_yaw_variance = float(rt_rp.get("mag_yaw_variance", mag_yaw_variance))
+        physical_gnss_observation_rate_hz = float(
+            rt_rp.get(
+                "gnss_profile_rate_hz",
+                physical_gnss_observation_rate_hz,
+            )
+        )
         tool_width = float(rt_rp.get("tool_width", tool_width))
         headland_width = float(rt_rp.get("headland_width", headland_width))
         num_headland_passes = int(rt_rp.get(
@@ -1146,6 +1157,7 @@ def generate_launch_description() -> LaunchDescription:
              "enable_mag_cal": enable_mag_cal,
              "mag_calibration_path": mag_cal_path,
              "stationary_seed_rate_hz": cog_stationary_rate,
+             "physical_gnss_observation_rate_hz": physical_gnss_observation_rate_hz,
              # Stationary-yaw aging penalty. The republish_latched path
              # adds (rate · age)² to the variance to model the chance
              # that the latched yaw has gone stale (manual rotation,
--- a/ros2/src/mowgli_bringup/test/test_launch_injection.py
+++ b/ros2/src/mowgli_bringup/test/test_launch_injection.py
@@ -108,8 +108,55 @@ def _node_parameter_keys(call: ast.Call) -> List[str]:
     return keys
 
 
+def _node_parameter_values(call: ast.Call, key: str) -> List[ast.expr]:
+    """Values assigned to one string key in a Node's parameter dicts."""
+    values: List[ast.expr] = []
+    for kw in call.keywords:
+        if kw.arg != "parameters" or not isinstance(kw.value, ast.List):
+            continue
+        for elt in kw.value.elts:
+            if not isinstance(elt, ast.Dict):
+                continue
+            for dict_key, dict_value in zip(elt.keys, elt.values):
+                if (
+                    isinstance(dict_key, ast.Constant)
+                    and dict_key.value == key
+                ):
+                    values.append(dict_value)
+    return values
+
+
 # ---------------------------------------------------------------------------
-# (a) num_headland_passes must reach coverage_server UNCLAMPED (issue #429).
+# (a) physical GNSS cadence must reach cog_to_imu (MGNSS-011).
+# ---------------------------------------------------------------------------
+
+
+def test_navigation_launch_wires_physical_gnss_rate_to_cog() -> None:
+    tree = _parse("navigation.launch.py")
+    assert _reads_robot_param(
+        tree,
+        "physical_gnss_observation_rate_hz",
+        "gnss_profile_rate_hz",
+    ), (
+        "navigation.launch.py must read the configured receiver-profile rate; "
+        "COG timing must not invent or infer a ROS publication rate."
+    )
+
+    call = _find_node_call(tree, "cog_to_imu")
+    assert call is not None, "navigation.launch.py no longer launches cog_to_imu"
+    values = _node_parameter_values(call, "physical_gnss_observation_rate_hz")
+    assert len(values) == 1
+    assert (
+        isinstance(values[0], ast.Name)
+        and values[0].id == "physical_gnss_observation_rate_hz"
+    ), (
+        "cog_to_imu must receive the bare physical GNSS rate loaded from "
+        "gnss_profile_rate_hz, not a publication-rate expression."
+    )
+
+
+# ---------------------------------------------------------------------------
+# (b) num_headland_passes must reach coverage_server UNCLAMPED (issue #429).
 # ---------------------------------------------------------------------------
 
 
@@ -162,7 +209,7 @@ def test_navigation_launch_does_not_clamp_num_headland_passes() -> None:
 
 
 # ---------------------------------------------------------------------------
-# (b) mowing_enabled must reach hardware_bridge_node (issue #195).
+# (c) mowing_enabled must reach hardware_bridge_node (issue #195).
 # ---------------------------------------------------------------------------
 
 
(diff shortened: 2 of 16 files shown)


cedbossneo:
## Review note: `status_pairing_window_s` default looks ~100× larger than what it guards

Recording the analysis behind the two behavioural changes in the PR body, because one of them touches the docking control path and the default is worth a deliberate decision before merge.

### Why the output rate becomes the position rate (this part is right)

`navsat_status_association.cpp:141-154`. When a status arrives carrying the same `position_observation_sequence` as the previous one but a later receipt stamp — the enrichment case, where the receiver merged GSA/GSV/health between two fixes — the receipt is closed and nothing is emitted. The code says it plainly: *"one output per PHYSICAL observation is the point of the whole gate"*.

So `/gps/absolute_pose` follows the **physical observation rate** (5 Hz) rather than the publication rate (10 Hz). That is a fix, not a regression: the previous behaviour republished the same physical position under two different timestamps, i.e. it manufactured observations that never happened. No argument here.

### Why the 0.25 s exists

`navsat_to_absolute_pose_node.cpp:130`, `status_pairing_window_s` default `0.25`. Note this is **not** a wait for the partner message — complete pairs are deliberately withheld until the window closes, so that a late *second* status carrying a conflicting sequence on the same receipt can still make the pair fail closed before anything is published (the `kAmbiguousReceipt` path).

### The concern

The skew this window covers is tiny. `/gps/fix` and `/gps/status` are built from **one** snapshot in `BuildSnapshot()` and published back to back, so the real inter-topic delay is DDS jitter — sub-millisecond to a few ms. 0.25 s is two orders of magnitude beyond that, and it is paid on every healthy sample to guard a case that only occurs on an upstream fault.

That cost is not confined to display. `/gps/absolute_pose` feeds `gps_dock_detection_node`, and `use_gps_dock_detection` defaults to `true`, so this is a live control input during the dock approach. Its own header documents the geometry:

```
p_dock_odom = p_robot_odom + R(-D)*(p_dock_map - p_robot_gps)
```

`p_robot_odom` is live; `p_robot_gps` now lags 0.25 s. Approaching the dock, the lagged GPS position sits *behind* the true one, so `(p_dock_map - p_robot_gps)` overstates the remaining distance by exactly the lag distance and the dock is computed **further away than it is** — the robot drives that much too far. At `min_approach_linear_velocity: 0.16` m/s that is ~4 cm, and more earlier in the approach where `approach_velocity_scaling_dist: 0.6` still allows higher speed.

Injecting a speed-dependent systematic offset into a computation whose entire header comment exists to keep the dock target centimetre-accurate is the wrong trade, especially with #515 (contact but no charge) still open on this same path.

### Suggestion

Lower the default to ~`0.05` s. That is still 10–50× the real skew, keeps the fail-closed guarantee for a genuinely conflicting late status, and cuts the docking-path error to under a centimetre. The parameter is already declared and tunable, so this is a one-line default change — it does not touch the association logic, which I think is correct as written.

Worth a second opinion on whether the withhold is needed at all on the healthy path, versus publishing immediately and reporting a late conflicting status as the receiver fault it is. That would be a design change rather than a default change, so not proposing it here.

PR #553, opened 2026-09-06
feat(gui): show host CPU usage alongside temperature

wjcloudy:
Diagnostics already shows CPU temperature, but does not show CPU utilization. Add host-wide CPU usage beside temperature in the summary badges and CPU card, using the existing ten-second diagnostics refresh.

The Go sampler reads aggregate Linux `/proc/stat` deltas and returns 0–100% across all cores. It shares samples across concurrent requests without sleeping or starting a background process. Warm-up, missing/malformed data, counter resets and long sampling gaps are shown as unavailable rather than 0%. The existing temperature API, health verdict and mower behavior are unchanged. Includes English/French labels and sampling documentation.

Validation: full Go backend suite, CPU sampler race tests, TypeScript, lint (0 errors), all 373 frontend tests across 48 files on Node 22, and the production frontend build pass. Confirmed read-only on a Raspberry Pi that the existing temperature endpoint works and its GUI container exposes host CPU counters. Upstream CI is running. No runtime changes were deployed to the mower.


cedbossneo (approved):
Checked the two things that usually go wrong with this kind of sampler. The sampler is created in the `getDiagnosticsSnapshot` factory, outside the returned handler, so it is one shared instance and not a fresh (always-nil) one per request. And the iowait-can-decrease case is handled by rebaselining rather than underflowing into a fabricated percentage. Reads `/proc/stat` directly with no external command, no extra privileges, informational only, and nil rather than a false zero when unavailable. Approving.

PR #554, opened 2026-09-06
fix(map): return obstacles only with their owning area

wjcloudy:
## Summary

With three mowing areas, one dig could appear as three overlapping obstacles in the GUI. `get_mowing_area` appended the global keepout cache to every area's response, giving copies in other areas a new anonymous tracker identity and losing their pending status.

Return each area's own obstacle entries, which already contain pending digs and promoted observations with their original metadata.

**Safety-critical review:** this response also feeds coverage planning. The global keepout cache and navigation mask are unchanged; tests verify the original dig location remains lethal before and after acceptance. No drive, blade, firmware or detection changes.

## Changes

- Remove the global-cache append from `get_mowing_area`; preserve owning-area IDs, names, sources and pending flags.
- Add three multi-area regression tests covering pending/discarded digs, accepted digs before and after reload, and promoted polygons. Include three lawns and an overlapping navigation area.
- Document the ownership/cache distinction in the handler.

Expiry policy and GUI accept/discard controls are outside this PR.

## Component

- [x] ROS2 Stack (`ros2/`)

## Testing

- [x] Built successfully: `mowgli_interfaces` and `mowgli_map` in ROS Kilted, as a non-root user.
- [x] Unit tests pass: all 68 C++ tests across the four map-package suites, including the three new regressions.
- All three new tests failed against the unmodified implementation, then passed with this fix.
- `colcon test --packages-select mowgli_map --return-code-on-test-failure`: 94 reported checks, zero errors/failures, 17 skipped (ament disables cppcheck 2.13.0 due to known performance issues).
- Changed lines verified with clang-format 18.1.3 against the merge-base with `origin/main`; `git diff --check` passed.
- Not deployed to a mower or tested in simulation. Full workspace validation remains for CI.

## Checklist

- [x] No unrelated changes
- [x] Physical behavior impact flagged for safety-critical review above
- [x] Changed C++ lines are clang-format 18 clean

### DIFF
--- a/ros2/src/mowgli_map/src/area_manager.cpp
+++ b/ros2/src/mowgli_map/src/area_manager.cpp
@@ -593,34 +593,19 @@ void MapServerNode::on_get_mowing_area(
       res->area.obstacle_info.push_back(info);
     }
 
-    // Also include persistent tracked obstacles from the obstacle tracker
-    // so the coverage planner can avoid them in the initial plan. Skip the
-    // ones already listed above: apply_promoted_obstacle writes a promoted
-    // keepout into BOTH stores, so without this the same hole was handed to
-    // F2C twice — and with obstacle_info it would also arrive a second time
-    // wearing the wrong identity (a pending dig looking like an accepted
-    // tracker obstacle).
-    const auto n_static = res->area.obstacles.size();
-    for (const auto& obs_poly : obstacle_polygons_)
-    {
-      if (obs_poly.points.size() >= 3 &&
-          !has_duplicate_obstacle(res->area.obstacles, obs_poly, kObstacleDedupEpsilonM))
-      {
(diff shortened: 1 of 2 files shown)


cedbossneo (approved):
The one thing that could make this dangerous is if the removed loop was the only route by which an un-promoted tracker obstacle reached the coverage planner. Checked every writer of `obstacle_polygons_`: it is written only by the load path (`area_manager.cpp:163`, which pushes into `entry.obstacles` in the same block) and by `apply_promoted_obstacle` (`progress_tracker.cpp:310`, which also checks `areas_[area_index].obstacles`). Nothing ever puts a non-promoted polygon in that cache, so every entry already lives in an area and the loop was pure cross-area leakage. The keepout mask remains global. Three tests pin ownership and provenance across reload. Approving.

PR #555, opened 2026-09-06
fix(gui): clarify vertex deletion and restore saved map rotation

wjcloudy:
## Summary

Deleting selected map vertices correctly removed points, but the confirmation claimed the entire area would be deleted. Choose point-specific confirmation text in Draw's direct-select mode, retaining the existing area warning for whole-feature selection. The point warning also explains that removing too many vertices can remove an invalid polygon.

A saved display bearing could appear in the rotation control while the map returned north-up: the bearing effect could run before the map loaded, and setting the map ref did not rerun it. Restore the bearing on map load as well as when the saved value arrives later.

## Changes

- Add English and French vertex-delete confirmation text, selected according to Draw's active mode (including a newly inserted midpoint whose public selection cache is not yet updated).
- Share camera synchronization between the full map and the compact dashboard map.
- Restore the bearing on load/replacement, remove old rotation listeners, and persist user gestures through the latest callback. Programmatic restoration does not trigger another save.
- Add 10 focused tests covering delete confirmations, load ordering, replacement/cleanup, and gesture versus programmatic rotation.

The existing GUI bearing setting and debounced persistence remain in use. This changes display rotation only; robot coordinates, map geometry, and mower motion are unaffected.

## Component

- [x] GUI (`gui/`)

## Testing

- [x] Built successfully: TypeScript check and Vite production build.
- [x] Unit tests pass: 383 tests across 50 files; final affected-test rerun 10/10 passed.
- ESLint repository target passed with zero errors; new camera and test files also checked with zero warnings.
- Translation key parity and `git diff --check` passed.
- Not deployed to a mower; no hardware or browser E2E run.

## Checklist

- [x] No unrelated changes
- [x] Physical behavior unaffected

### DIFF
--- a/gui/web/src/i18n/locales/en.json
+++ b/gui/web/src/i18n/locales/en.json
@@ -1581,6 +1581,8 @@
     "selectExactlyOneToSplit": "Select exactly 1 area to split",
     "selectionMustBePolygon": "Selection must be a polygon area",
     "drawLineAcrossArea": "Draw a line across the area to split it",
+    "deletePointsConfirmTitle": "Delete selected points?",
+    "deletePointsConfirmBody": "The selected points will be removed. If too few points remain to form a polygon, that shape will also be removed. You can undo before saving.",
     "deleteAreaConfirmTitle": "Delete this area?",
     "deleteAreaConfirmBody": "The selected area (and its obstacles) will be removed. You can undo before saving.",
     "deleteAreaConfirmOk": "Delete",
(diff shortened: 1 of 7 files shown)


cedbossneo (approved):
Two distinct fixes and both read correctly: the bearing camera is restored regardless of whether config or the map instance arrives first (the ref-set-doesn't-rerun-effect trap is handled by calling `setBearing` in `onLoad`), and the trash action now distinguishes `direct_select` (vertices) from `simple_select` (whole features) so the confirmation dialog says what will actually happen. Listener detach on unmount is preserved. Approving.

PR #556, opened 2026-09-06
fix(hardware): publish blade ESC current in amperes

wjcloudy:
## Summary

Blade current is published 1,000 times too high: a Yardforce ESC reading of 611 mA becomes `Status.mower_esc_current = 611` and the GUI displays 611 A. Convert milliamps to amperes at the ROS2 hardware bridge, giving 0.611 A. Physical behavior is unaffected.

## Changes

- Convert blade telemetry through a small tested helper before publishing ROS status, so GUI, diagnostics and MQTT receive amperes.
- Correct both protocol-header comments and the firmware wiki: `power_watts` is a misleading legacy name for raw ESC current in mA. Preserve the field, packet layout and protocol version.
- Add regression coverage for a literal 16-byte packet containing 611 mA, plus zero, 1 mA, 600 mA and the uint16 maximum.

### Unit evidence

Both Yardforce 500 and 500B use the [same ESC decoder](https://github.com/mowglinext/mowglinext/blob/3834fb810de777a0fd956a027fdd1901801ddcac/firmware/stm32/ros_usbnode/src/blademotor.c#L282), which reads UART bytes 9–10 into `BLADEMOTOR_u16Power`; the [firmware sender](https://github.com/mowglinext/mowglinext/blob/3834fb810de777a0fd956a027fdd1901801ddcac/firmware/stm32/ros_usbnode/src/ros/ros_custom/cpp_main.cpp#L1437) forwards that integer unchanged. Neither computes watts. The [ROS1 bridge explicitly divided the same value by 1,000](https://github.com/Nekraus/Mowgli/blob/50f31f603131e9da646ea412b719178ea6d667b5/stm32/ros_usbnode/src/ros/ros_custom/cpp_main.cpp#L598) for ESC current.

The same missing conversion is present on upstream dev/main and release tags v1.0.2 through v1.1.0. The simulator already publishes amperes directly and needs no change. This restores the existing firmware's interpretation; no reflash is needed.

## Component

- [x] ROS2 Stack (`ros2/`)
- [x] Firmware (`firmware/`) — comment only
- [x] Documentation (`wiki/`)

## Testing

- [x] Built `mowgli_hardware` against ROS Kilted as a non-root user.
- [x] Package tests: 247 reported checks, zero errors/failures; 44 cppcheck skips due to the installed cppcheck 2.13 performance exclusion. All 34 protocol tests pass, including both new tests.
- [x] clang-format 18 changed-line check against the merge-base with origin/main.
- [x] Protocol-version guard: v6 fingerprint unchanged, host and firmware agree.
- [x] `git diff --check`.
- [ ] Deployed fix on hardware — live pre-fix telemetry confirmed the 611 reading; this PR has not been deployed.

## Checklist

- [x] Targets `dev` and follows conventional commit format.
- [x] Documentation updated; no secrets or unrelated changes.
- [x] Physical behavior (blades, motion, e-stop and safety gates) unaffected.
- [x] C++ changed lines are clang-format 18 clean.
- Protocol headers have comment-only changes: no wire change or version bump.
- ROS interfaces and configuration schemas are unchanged.

### DIFF
--- a/firmware/stm32/ros_usbnode/include/mowgli_protocol.h
+++ b/firmware/stm32/ros_usbnode/include/mowgli_protocol.h
@@ -595,7 +595,7 @@ typedef struct {
   uint8_t type;         /**< PKT_ID_BLADE_STATUS */
   uint8_t is_active;    /**< 1=running, 0=stopped */
   uint16_t rpm;         /**< Blade motor RPM */
-  uint16_t power_watts; /**< Power consumption [W] */
+  uint16_t power_watts; /**< ESC current [mA]; legacy field name, not watts */
   float temperature;    /**< Blade/motor temperature [C] */
   uint32_t error_count; /**< Cumulative error counter */
   uint16_t crc;         /**< CRC-16 CCITT over preceding bytes */
--- a/ros2/src/mowgli_hardware/include/mowgli_hardware/blade_telemetry.hpp
+++ b/ros2/src/mowgli_hardware/include/mowgli_hardware/blade_telemetry.hpp
@@ -1,0 +1,22 @@
+// Copyright 2026 Mowgli Project
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#ifndef MOWGLI_HARDWARE__BLADE_TELEMETRY_HPP_
+#define MOWGLI_HARDWARE__BLADE_TELEMETRY_HPP_
+
+#include "mowgli_hardware/ll_datatypes.hpp"
+
+namespace mowgli_hardware
+{
+
+constexpr float blade_current_amps(const LlBladeStatus& packet)
+{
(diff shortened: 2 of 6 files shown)


cedbossneo (approved):
Verified the core claim against the firmware rather than taking it on faith: `blademotor.c:282` sets `BLADEMOTOR_u16Power = data[9] + (data[10]<<8)` — the raw ESC UART bytes 9-10, little-endian, exactly as described — and `cpp_main.cpp:1434` forwards it unchanged into `power_watts`. Every existing consumer already assumed amperes: `diagnostics_node` publishes the key as `mower_esc_current_a` and the sim bridge emits 0.5. So this was a real ×1000 unit bug, not a reinterpretation. No consumer thresholds on this value, wire layout is untouched, and the firmware change is comment-only. Approving.

PR #557, opened 2026-09-06
fix(gui): preserve NTRIP toggle state after saving

wjcloudy:
## Summary

After disabling NTRIP and saving, the correction toggle can appear enabled again on reload even though the NTRIP client is stopped. Saving `ntrip_enabled: false` prunes the schema-default value from the sparse YAML and persists `GNSS_NTRIP_ENABLED=false` in the runtime environment. The settings API then returns that fallback as the string `"false"`, which the React switch treats as truthy.

Return JSON booleans for NTRIP and GGA environment fallbacks so the displayed state agrees with the saved configuration. Physical behavior is unaffected: this changes settings serialization, not receiver configuration or the correction client.

## Changes

- Convert `GNSS_NTRIP_ENABLED` and `GNSS_NTRIP_GGA_ENABLED` fallbacks to booleans using the same accepted spellings as `start_gps.sh`.
- Preserve explicit YAML precedence and existing numeric/string fallback handling.
- Add API tests for boolean spellings and a full save-off → default pruning → reload regression.

## Component

- [x] GUI (`gui/`)

## Testing

- [x] `go test ./pkg/api -count=1` passes in a non-root Go 1.24 container.
- [x] Changed files formatted with `gofmt`; `git diff --check` passes.
- [x] Pre-fix behavior confirmed on a running mower: the API returned `ntrip_enabled: "false"`, its GPS container had `GNSS_NTRIP_ENABLED=false`, and no `ntrip_node` was running.
- [ ] Fix deployed on hardware — not yet deployed.

## Checklist

- [x] Targets `dev`; conventional commit format.
- [x] No hardcoded secrets or unrelated changes.
- [x] Physical behavior (blades, motion, e-stop and safety gates) unaffected.
- No ROS interfaces, COBS wire format, configuration schema or documentation changes required.

### DIFF
--- a/gui/pkg/api/gnss_runtime_config.go
+++ b/gui/pkg/api/gnss_runtime_config.go
@@ -218,6 +218,18 @@ func applyGNSSRuntimeFallback(flat map[string]any, explicitFlat map[string]any, 
 			return
 		}
 	}
+	if envKey == "GNSS_NTRIP_ENABLED" || envKey == "GNSS_NTRIP_GGA_ENABLED" {
+		// Environment values are strings, but settings switches require JSON
+		// booleans: the string "false" is truthy in JavaScript. Match the GPS
+		// startup script's normalize_bool, including its accepted spellings.
+		switch strings.ToLower(envValue) {
+		case "1", "true", "yes", "y", "on":
+			flat[targetKey] = true
+		default:
+			flat[targetKey] = false
+		}
+		return
+	}
 	flat[targetKey] = envValue
 }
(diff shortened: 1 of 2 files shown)


cedbossneo (approved):
Correct diagnosis: env values are strings, the string "false" is truthy in JS, and the accepted spellings match `normalize_bool` in the GPS start script. The round-trip test (disable → prune from YAML → env written → reload reads false) is exactly the regression that matters. Approving.

PR #558, opened 2026-09-06
feat: add automatic blade direction and status display

wjcloudy:
## Summary

Add **Automatic blade direction** in Settings → Mowing, consistent Forward/Reverse/OFF map controls, and a **Requested blade direction** display in diagnostics and the dashboard.

The feature lets compatible mowers select either cutting direction without the behavior tree overwriting an operator's choice. It makes command intent visible while preserving the distinction between a requested direction and actual physical rotation.

## Supported behavior

- Automatic direction defaults to **off**. When enabled, the first permitted blade-enable request randomly chooses Forward or Reverse. The choice persists through pauses, area changes, temporary stops and recharge; consecutive sessions can choose the same direction. Changing the setting requires a ROS2 restart.
- Map **Blade Forward / Blade Backward / Blade Off** controls work on desktop and mobile. A named direction overrides the automatic choice; selecting it while idle deliberately preselects the next permitted enable without starting a mission. Idle, transit, docking and guard OFF decisions retain precedence.
- **Blade Off** inhibits subsequent tree enables. Explicit Play/Resume, Mow next area, Mow this area or Manual Mowing clears only that inhibition and retains the selected direction. Automatic continuation does not clear it. EndSession resets the session policy. Software OFF is not isolation for handling the mower.
- Optional lift recovery preserves the latest requested direction and its delay. Later OFF, emergency/STOP, dry-run suppression and lost serial communication cancel pending recovery; repeated ON cannot bypass the delay, and a second lift restarts it.
- Diagnostics and dashboard show **Forward, Reverse, Off or Unknown** for the last command successfully written to the firmware link. RPM stays separate. Failed writes/reconnect clear the display; missing/older data and a stopped status stream show Unknown. Firmware help is visible in diagnostics and available by touch on the dashboard.

## How it works

The behavior tree owns one in-memory blade policy. Its tick, operator service and start handlers share an explicit mutually exclusive callback group. Coverage, manual mowing and operator requests also share one ROS client, preserving their request order.

A new operator `BladeControl` service reports policy acceptance, queued forwarding and a reason. The existing hardware `MowerControl` service and STM32 wire protocol remain unchanged; the earlier BT endpoint is a thin compatibility adapter.

For OFF, the GUI first attempts the BT latch with a 250 ms budget, then independently requests hardware OFF with a two-second budget. It still attempts OFF if the BT is absent, rejects the request, stalls or the browser closes. Partial success is shown as a warning: a direct OFF without the latch may be overwritten by an older tree; a retained latch without successful hardware contact does not prove physical stoppage. There is **no direct hardware ON fallback**. Direction is validated only for ON.

## Firmware, deployment and safety

**Safety-critical: affects requested blade rotation.** Use reversal only with a bidirectional blade assembly and firmware that waits for the blade to stop before reversing. Firmware remains the safety authority. Stock/older firmware may ignore Reverse and still run forward; the displayed value is a command, not signed RPM or a firmware acknowledgment.

Land/release firmware companion **#559 before #558**, and flash compatible firmware before enabling reversal. This PR does not infer capability from version numbers, implement a host-side reversal dwell, or include LFP charging changes.

The added ROS `Status` field requires compatible rebuilt ROS consumers and GUI. The new GUI reports an unavailable/older operator service clearly, retains the compatible direct OFF path with an explicit warning, and never bypasses the tree for ON. Update ROS and GUI together.

## Validation

- Affected Kilted packages build as the normal user. Behavior, hardware and monitoring suites: **693 reported checks, zero errors/failures, 109 skipped** (424 / 236 / 33 by package).
- Real service/timer tests cover all explicit start entry points, unchanged session direction, a blocked production BT tick under MultiThreadedExecutor, OFF during a bridge outage, invalid OFF direction and pending ON→OFF request ordering.
- Pseudo-terminal tests inspect actual outgoing blade packets for forward/reverse lift recovery, newer direction choices, repeated ON, OFF during lift/delay, repeated lift, STOP/emergency, dry-run and disconnect/failed sends.
- Go API/provider/generated-message suites pass, including independently failing services, bounded OFF fallback and browser cancellation. Frontend unit/i18n checks: **19 passed**. **12 distinct mocked Chromium scenarios** pass across desktop/mobile, including visible compatibility warnings and touch-accessible firmware help. TypeScript, production build and changed-source ESLint pass (five existing MowerActions warnings).
- Ten launch/config tests pass. Configuration, firmware protocol and generated rosserial-header guards pass; Go bindings regenerated. Changed C++ passes clang-format 18 against the origin/main merge-base; `git diff --check` passes.
- **HARDWARE_REQUIRED for these review fixes.** Earlier 500/500B physical rotation reports and their exact baselines remain recorded in #559. They do not establish the new recovery behavior or complete stop-before-reverse/emergency/missing-feedback acceptance. The baseline-recording requirements, supervised blades-removed procedure and pass/fail criteria are in [the feature documentation](https://github.com/wjcloudy/mowglinext/blob/4498cde3548116a9aa375f220cb35f36776205bd/docs/BLADE_AUTO_DIRECTION.md#validation-before-enabling-on-a-mower). No mower was deployed or commanded for this revision.

## Screenshots

Mocked telemetry and service outcomes, not evidence of physical rotation. These assets live on a separate evidence branch and are not part of the product screenshot directory.

### Diagnostics: requested direction and firmware requirement

Diagnostics: requested direction and firmware requirement — desktop

Diagnostics: requested direction and firmware requirement — mobile

### Dashboard: tap the requested direction for firmware help

Dashboard: tap the requested direction for firmware help — desktop

Dashboard: tap the requested direction for firmware help — mobile

### Map: hardware OFF accepted but persistent inhibition unavailable

Map: hardware OFF accepted but persistent inhibition unavailable — desktop

Map: hardware OFF accepted but persistent inhibition unavailable — mobile


cedbossneo (changes requested):
Thorough review — the design here is genuinely good and most of the safety story holds up. One blocking regression, then nits.

## What I verified as correct (not up for re-litigation)

- **Wire protocol unchanged.** `send_blade_command` only captures the existing `send_raw_packet` return value; `pkt.blade_dir` already existed. Framing, emergency path and the dry-run gate are untouched, and every operator command still funnels through `on_mower_control`.
- **`hardware_bridge` only records intent**, after a *successful* write, and clears it on `reset_serial_dependent_state()`. Correctly separated from RPM/activity.
- **An operator ON can never enable the blade against the tree.** `forOperatorCommand` returns `enabled = requested_enabled_ && !operator_inhibit_`, so operator input can only ever *reduce* motion. `forCommand` never selects or changes direction on an OFF. The map ON path is strictly tighter than what ships today.
- **No callback race.** `tick_timer_` is a plain `create_wall_timer` and the service uses no explicit group, so both sit in the node's default mutually-exclusive group — the "MUST share the owner's group" comment holds.
- **Go routing is right.** `blade_control` always attempts a direct hardware OFF (bounded 2 s) before the BT latch, and never sends a direct ON.
- **Invariant 15 satisfied.** Default `false` lives only in the in-package template; the GUI schema default matches; the installed sparse config is untouched.

---

## [BLOCKING] The operator OFF-inhibit survives into the next mow, which then runs blade-off end to end

`blade_direction.hpp:69` latches `operator_inhibit_ = !enabled`. The **only** reset is `endSession()`, called solely from `EndSession` (`status_nodes.cpp:171`). In `main_tree.xml`, `EndSession` appears at :462, :1027, :1044, :1089, :1137, :1193 only — completed-and-docked, home-and-docked, charger-failed, recording-complete. It is **not** in `IdleSequence`, **not** in `StopHoldSequence`, and `COMMAND_START` clears only `clearSingleAreaMode` (`behavior_tree_node.cpp:622-625`). Lines 411 and 469 of the tree document further paths that deliberately fall through *without* `EndSession` (low-battery dock + auto-resume), so a recharge cycle keeps it too.

Meanwhile the three blade menu items render unconditionally (`MowerActions.tsx:158-183`) — including while docked and idle.

**Failure scenario.** Robot is docked and idle. The operator taps **Blade off** — the natural "make sure that thing can't spin" reflex before touching the machine. `requested_enabled_` is already false and the bridge is reachable, so the service returns `success = true` and silently latches `operator_inhibit_ = true`. Nothing in the UI reflects that a latch was set. The operator later taps Play. `FollowStrip::setBladeEnabled(true)` → `forMowerCommand` → `effective = false` → `{0, 0}` goes on the wire for the **entire mission**. The robot drives the full coverage plan with the blade never spinning, and because `map_server`'s blade gate refuses to stamp progress without a spinning blade, coverage % never advances either. The mow just looks broken.

**This is a regression, not a pre-existing trap.** Today "Blade off" is a one-shot direct write to the bridge that the tree's next `SetMowerEnabled` tick simply overrides. This PR converts it into sticky session state with a much longer lifetime than the action that set it.

Recovery exists but is undiscoverable: tapping **Blade forward** clears the latch (`forOperatorCommand(true, …)` sets `operator_inhibit_ = false`).

**Suggested fix**, in the `COMMAND_START` branch at `behavior_tree_node.cpp:622-625`, right next to `clearSingleAreaMode`: clear the operator inhibit when a new mow is commanded. Please keep it a *targeted* clear rather than a full `endSession()` there — `endSession()` would also reset `direction_`, and re-rolling the random direction on every `COMMAND_START` would change the "one direction per session across pauses and recharge" property the PR is built around.

Whichever way you cut it, the invariant worth stating in a comment is: **an operator inhibit must not outlive the mission it was issued during.**

---

## Nits (non-blocking)

- `blade_direction.hpp:71` — `forOperatorCommand(true, dir)` pins `direction_` even when `requested_enabled_` is false, so tapping "Blade reverse" while idle pre-loads reverse for the next session start. Defensible, but worth a comment saying it is deliberate.
- `coverage_nodes.cpp:1120` — `forMowerCommand` is called *before* the `wait_for_service(200ms)` early return, so the session direction is latched even when the request is never sent. The `forMowerCommand` comment says recording an undelivered OFF is intentional; a one-line note at the call site would stop the next reader treating it as a bug.
- `blade_control_service.hpp` — the early `return`s leave `resp->success` at its default `false` with an empty `message`. The PR body promises the operator sees "the server's rejection reason"; populating `message` on the reject paths would make that true.
- `docs/screenshots/` is documented (its own README) as the README/wiki/Pages asset set. The two blade-direction shots are PR evidence and would sit better on the PR itself.

---

## Verdict

**Request changes** — on the blocking item alone. Everything else here I am happy with, and the safety architecture (operator can only inhibit, never enable; firmware remains the authority) is the right shape.

Separately, and not a blocker for this PR: with #559 unmerged the firmware ignores the direction request on 500/500B, so `blade_auto_reverse` is inert. Given both defaults are `false`, merging this alone is harmless — but enabling the pair needs the field check with blades removed that #559's checklist still has unticked.

cedbossneo:
Follow-up to my review above — a second pass turned up more, including one item that belongs in the same fix batch as the OFF latch. Grouping them here so you can address everything in one go.

### 2. `hardware_bridge_node.cpp:1530` — lift recovery hardcodes forward, which becomes a direction change on a spinning blade

`send_blade_command(1, 0)` re-enables the blade with direction 0 regardless of the session's direction. The line predates this PR and was harmless while direction was always 0; it stops being harmless the moment a session can be in reverse.

With `lift_recovery_mode:=true` (parameter default is `false`, `hardware_bridge_node.cpp:656`, so this needs opting in) and a reverse session: the bridge spins the blade **forward** `lift_blade_resume_delay_sec` after the lift clears, then the tree re-asserts `{1, 1}` on its next tick — 10 Hz during manual mowing — i.e. a reversal request roughly 100 ms after spin-up, far inside spin-down.

The firmware stopped-reversal guard is the thing that saves this, and per Invariant 1 that is exactly where the authority belongs — so this is not a safety hole. It is still wrong: the bridge should re-send the **last requested direction** rather than assume forward, otherwise it fights the session policy this PR just introduced.

### 3. The tile can report "Reverse" while the blade physically spins forward

`blademotor.c:246-263` has the reverse branch commented out on every firmware in this repo — `direction` is latched and discarded. So on stock firmware "Blade backward" reports **Reverse** in the new display while the shaft turns forward. The PR is careful to say the field is command intent, not measured rotation, and I think that framing is right — but a tile that says "Reverse" on hardware that cannot reverse is a trap during commissioning. Worth gating the reverse UI on a firmware capability signal, or at minimum saying so in the tile's help text.

Related sequencing point: once #559 lands, `blade_control_service.hpp` forwards `{1, new_direction}` with no OFF, dwell or capability gate of its own. That is fine **if** #559 merges first and owns the guard. Please land them in that order.

### 4. `blade_control_service.hpp:41-42` — an OFF is rejected on a field the OFF does not use

```cpp
if (!ctx || req->mow_enabled > 1u || req->mow_direction > 1u)
  return;
```

Two lines above the comment "Always latch OFF, including during a hardware-bridge outage", a stop request carrying a garbage `mow_direction` is dropped entirely. Not reachable from the GUI (`mowglinext.go` clamps both fields), so this is defensive only — but a stop should never be refused because of a field it ignores. Validate direction only on the enable path.

### 5. `mowglinext.go:605-607` — new GUI + old ROS image breaks the blade menu

`/behavior_tree_node/mower_control` does not exist on an older ROS image, so Forward/Backward fail outright, and OFF physically succeeds (the direct hardware attempt runs first) yet still surfaces an error to the operator. Since the GUI and ROS containers update independently, this is a realistic intermediate state. Either tolerate a missing BT service on the OFF path, or call it out in the compatibility section — the PR currently says only "rebuild/deploy ROS consumers and the GUI together".

### 6. `bt_context.hpp:116-118` — new shared state opts out of the documented `context_mutex` contract

`blade_direction` is mutated from both the service callback and the tick, and is safe **only** because both land in the default mutually-exclusive callback group. That is asserted in a comment with no test behind it. It is a live trap: giving the service a reentrant group — the natural fix if someone later trips over the 1 s `wait_for_service` in `SetMowerEnabled` — silently introduces a race on blade-enable state. Either take `context_mutex` like the surrounding state does, or add a test that pins the callback-group assumption.

### 7. `mowglinext.go:595-605` — direct OFF runs before the BT latch

The bounded direct-hardware OFF is sent first, then the latch. During manual mowing the tree re-asserts ON at 10 Hz, so the blade can re-spin for up to ~100 ms after the operator pressed OFF. Latching first and sending the direct OFF second closes it, and costs nothing.

---

To be clear about priorities: **only item 1 in my previous review blocks.** Items 2 and 4 I would like in the same batch since they are a few lines each and sit in the same code you will be touching. 3, 5, 6 and 7 are fine as follow-ups if you would rather keep this PR tight — say so and I will not hold the merge on them.

The core design — operator can only inhibit and never enable, firmware keeps the authority, intent reported separately from measured RPM — is the right architecture and I would like to get it in.

wjcloudy:
I have addressed all seven review points and the four original nits in **4498cde3548116a9aa375f220cb35f36776205bd**, after merging current dev (`9ebe2555`).

| Review point | Change and validation |
| --- | --- |
| **1. OFF inhibit surviving into the next mow** | Added a targeted inhibition clear for explicit START/S2, Manual Mowing and `start_in_area`. It neither resets direction nor sends ON. Automatic continuations keep inhibition. Production-service regressions exercise every entry point and preserve a preselected reverse direction. |
| **2. Lift recovery hardcoded Forward** | Recovery uses the latest desired direction, independently of the last-written display value. New OFF cancels delayed recovery; repeated ON cannot bypass its delay. Repeated lift, STOP/emergency, dry-run and serial loss invalidate/restart recovery appropriately. Pseudo-terminal tests check outgoing packets. |
| **3. Reverse display on unsupported firmware / merge order** | Added explicit firmware/assembly warnings in EN/FR, including touch-accessible dashboard help. The display remains requested intent; no capability is guessed from a version string. The PR/doc now explicitly require #559-before-#558 sequencing and compatible firmware before reversal use. |
| **4. OFF rejected for an unused direction** | Both Go and ROS validate direction only on ON. ROS tests cover OFF with 2/255, including a missing bridge; the API covers 255 and normalizes hardware OFF to direction 0. |
| **5. New GUI with older/unavailable ROS** | No direct ON fallback. OFF still uses the unchanged hardware service and returns a visible partial-success warning if the persistent latch is unavailable. The warning explicitly says the tree may re-enable it. Other combinations distinguish retained inhibition from unconfirmed hardware contact, and both failures are reported. |
| **6. Callback ownership** | Kept the mutually exclusive ownership model, made the group explicit for the tick/operator/start handlers, and corrected the context documentation. A real production tick is held under a three-thread executor and the test proves the operator callback cannot overlap it. No claim of a pre-existing race and no scattered context-mutex additions. |
| **7. Direct OFF preceding the latch** | Reordered to latch first, then independent direct OFF. The latch has a 250 ms budget and the fallback its own two seconds, including when the browser disconnects. BT blade paths share one client/writer; regression coverage checks pending ON then OFF ordering and repeated ticks after inhibition. This does not claim global ordering across unrelated DDS clients or replace emergency stop. |

The original nits are also covered:

- Idle Forward/Reverse preselection is documented as intentional and tested without idle blade enable.
- Both tree call sites explain why direction selection precedes discovery/send: it records intent and must not reroll on retry.
- `MowerControl.srv` actually had no message field. I added an operator-specific `BladeControl.srv` with acceptance, queued-forwarding status and explanatory text, regenerated bindings, and kept the hardware schema unchanged for the compatible OFF path. The legacy BT service is a thin adapter over the same policy.
- Removed the two PR-only images from `docs/screenshots`. Refreshed desktop/mobile evidence is shown directly in the PR using immutable links to a separate evidence branch, without collapsed sections.

Validation: affected ROS suites **693 checks, 0 errors/failures, 109 skipped**; Go API/provider/message suites; **19 frontend tests**; **12 distinct browser scenarios**; **10 launch/config tests**; TypeScript, production build, changed-source lint, codegen/protocol/config guards and clang-format 18 checks pass.

I have also updated the physical-evidence wording: #559 already records operator-confirmed forward/reverse operation for the specific 500 and combined 500B baselines. The new review fixes and the full stopped-reversal/stop/missing-feedback hardware procedure remain **HARDWARE_REQUIRED**, with the baseline requirements and acceptance procedure documented. No deployment or physical blade command was performed for this revision.

@cedbossneo could you please re-review? I have addressed all the points above, including the explicit-start inhibition lifetime and mixed-version OFF warning behavior. The latest follow-up (01ce7dc9) shortens that warning and refreshes both screenshots; its focused API/UI/browser tests pass. The full local ROS suite passed as recorded above. Current-head GUI tests, both Docker builds, firmware checks, formatting and static analysis pass; the current-head ROS CI rerun is still in progress. The documented hardware acceptance remains outstanding.

cedbossneo (commented):
Re-reviewed after the changes: the blocking regression from my previous review is fixed at the exact place and in the way requested, with tests; the four nits are addressed. Also verified: the `Status.h` change under `firmware/.../ros_lib` is regenerated dead code with no wire-protocol effect; `hardware_bridge` only records intent after a successful write and clears it on serial loss; lift recovery is narrower than before (and now honours dry-run); template/schema parity holds; CI green.

One cleanup pushed to this branch as `938712af` (maintainer edit): `behavior_tree_node` never served `~/mower_control` on `dev`, so the "compatibility for the earlier combined test GUI" adapter in `blade_control_service.hpp` was a second permanent endpoint reusing the hardware `MowerControl` schema and a real source of confusion with `/hardware_bridge/mower_control`. Removed, docs/codemaps updated; `~/blade_control` is the only blade endpoint on the tree node.

Follow-up issues rather than blockers: the operator "Blade Off" inhibition is invisible in the UI and survives automatic resumes; the map "Blade Backward" entry is not conditioned on firmware capability (without #559 the display says Reverse while the blade turns forward); the `Status.msg` field addition changes the message hash for any consumer not rebuilt.

Merge order: #559 first (firmware reverse), then this one. Approving once CI is green on `938712af`.

cedbossneo (approved):
CI green on ba810c23 (dead compatibility service removed, test harness now driven through ~/blade_control). Blocking regression from my earlier review is fixed; approving. Merge after #559 (firmware reverse) so the Reverse display matches physical rotation.

cedbossneo:
Merged current `dev` into your branch (d85b49c6) to get CI back on a fresh base. Four add/add conflicts, all kept both sides: the `blade_auto_reverse` key next to dev's new `blade_load_*` keys in `useSettingsManager.ts`, the `bladeDirection` / `systemPower` blocks in en/fr, and your two new hardware-bridge tests after dev's. No functional change; still sequenced after #559.

cedbossneo:
My dev merge broke the ROS2 build: the add/add resolution of `test_hardware_bridge_parameters.cpp` dropped the closing brace of `DryRunAndFailedSendCannotResume`, nesting dev's two cmd_vel slew tests inside it. Restored in the commit just pushed; `mowgli_hardware` builds against your updated `Status.msg` and passes 268 tests locally. Sorry for the noise — nothing else changed.

wjcloudy:
@cedbossneo I've pushed **90939258** to fix the blade-service test fixture after the latest dev merge: it now creates the helper node before registering services, as production initialization does. The newly added coverage-orientation service needs that node; its absence caused the test setup to crash. No production behavior changed in this follow-up.

Formatting, static analysis, GUI tests, both Docker builds and firmware checks pass at this head; the ROS build/test check is still running. The isolated local test attempt stopped at an older Nav2 dependency, so I am not counting that as a passing test run.

Your earlier blade review fixes remain in place. #559 now also includes external-IMU and onboard tilt-sensor bus recovery, with passing firmware CI and explicit hardware-validation limits. Once **#559's remaining physical acceptance is resolved and merged**, and this PR's current-head ROS CI passes, could you merge **#558 immediately afterward**, preserving the agreed firmware-first order?

PR #559, opened 2026-09-06
fix(firmware): blade direction guard and sensor bus recovery

wjcloudy:
## Summary

Honor explicit forward/reverse blade requests on Yardforce 500 and 500B firmware, with an OFF interval and controller-feedback guard before changing direction. Previously, the setter ignored direction and the frame builder always sent forward. This is the firmware companion to #558 and also supports existing explicit direction requests. It also recovers the external IMU and onboard tilt-sensor buses so an interrupted sensor transaction can be repaired without a power cycle.

**Safety-critical: changes physical blade behavior.** The standard 500 bench capture shows a held speed word followed by zero, not live spin-down RPM. Visual observation supports that run, but does not establish how the ESC detects stopping. Broader physical validation and equivalent 500B coast-down evidence remain outstanding.

## Behavior

- Send the requested direction using the existing packet; reverse is `C0` with checksum `62`, attributed to Jeremy Salwen's original fix.
- Before a direction change, send OFF and wait at least 1000 ms after an accepted transmission. Require valid, inactive, error-free zero-speed reports spanning at least 300 ms, with no reply gap greater than 300 ms.
- Release reversal only on a newly received qualifying reply. Held nonzero values never count toward zero confirmation, and a cached zero cannot release reversal when the OFF dwell expires.
- Keep sending OFF if feedback does not qualify, with throttled diagnostics after five seconds. There is no forced-start timeout.
- Cancel pending reversal on OFF or controller error. Preserve emergency, heartbeat and idle gates, DMA ownership, and receive polling while transmission is busy.

A separate `Yardforce500_COASTDOWN_VALIDATION` build supports forward/OFF tracing and always inhibits reversal. Release packaging excludes it. Charging profiles and USB packets remain unchanged. Sensor faults now inhibit both wheels and blade; no LFP-specific charging changes are included. Protocol remains 6; firmware versions are generated from the commit.

## Sensor recovery

- **External IMU (software I2C):** bring in the existing bounded clock/STOP recovery. Retry at most once per second, fail invalid reads, and retain the configured sensor; no blocking reprobe or fabricated samples.
- **Onboard LIS3DH (I2C1):** the main loop owns transactions and recovery. USB emergency callbacks consume a qualified snapshot, so they cannot block or compete for the bus. Failed/stale reads inhibit motion.
- Recover held SDA with up to nine open-drain clocks and STOP, then verify sensor identity and restore/read back its safety configuration. Polling and periodic configuration audits detect failures and sensor resets. Recovery is spread across main-loop passes; no MCU reset is required.
- Sensor faults remain latched after communication recovers. Only an explicit release with healthy, untriggered sensor status may clear them; heartbeat-only recovery cannot clear a sensor fault. Existing emergency indication and foreground diagnostics report the failure.

This addresses the observed case where a stuck onboard bus delayed the main loop and starved blade-status telemetry. [Design, regression coverage, timing limits and hardware acceptance](https://github.com/wjcloudy/mowglinext/blob/deefc5d2/firmware/stm32/ros_usbnode/I2C-RECOVERY.md).

The new automatic onboard recovery is **HARDWARE_PENDING**. The earlier manual bus-clear result validates that procedure on its recorded LFP baseline, not this new firmware implementation. It does not resolve the separate blade coast-down evidence limitation.

## Testing

- Both standard firmware builds and the validation build pass.
- Production-driver regression tests pass for all three configurations, including both directions, held-then-zero feedback, extended holds, interrupted confirmation, cached-zero rejection, invalid/missing feedback, cancellation, busy/failed transmission and tick wrap.
- External-IMU and onboard-sensor fault-injection tests pass, including the real heartbeat callback, held lines, failed reads/writes, configuration loss, stale snapshots and fault-latch preservation. These run in Firmware CI.
- Protocol, board-default and generated-header checks pass. Current-head Firmware CI passes all applicable jobs.
- Standard 500 capture at `962ce210` plus validation instrumentation: valid replies continued about every 100 ms after OFF. The speed word held at `3494`, then cleared at ~2.26 s; the operator observed the rotor stopped before the zero report. No progressive coast signal was observed in the additional response words.

The 1000 ms minimum is not a measured worst-case stopping time. Fresh UART replies do not prove fresh speed measurement inside the ESC. [Recorded evidence, earlier direction tests and remaining physical acceptance procedure](https://github.com/wjcloudy/mowglinext/blob/c942ba31/firmware/stm32/ros_usbnode/BLADE-REVERSE.md) retain the scope and provenance of each test.


cedbossneo (commented):
Review (firmware, blade-safety focus). Verified against `origin/dev`:

**Confirmed correct**
- Reverse frame `C0` + checksum `62` re-derived byte by byte with the driver's existing checksum routine; forward/OFF frames unchanged.
- Interlock does what the description says: OFF first, ≥1000 ms since an accepted OFF, then fresh/valid/inactive/0-RPM/error-free feedback for ≥300 ms; missing or invalid feedback keeps the blade OFF with no forced-start timeout. Pending reversal is cancelled by OFF; repeated ON does not bypass the guard.
- No existing gate is bypassed (emergency, heartbeat, IDLE, `on_cmd_blade`). DMA-owned buffer is not written. Native regression tests run in CI. Rebase on `dev` (20 commits behind) merges with no conflict. No regression for boards that never request reverse.

**Blocking before merge — HIGH**
The guard's safety rests on an unverified assumption: that `BLADEMOTOR_u16RPM` (reply bytes [7..8]) and `BLADEMOTOR_bActivated` (byte [5] & 0x80) reflect *measured* rotor motion rather than values the ESC zeroes as soon as it is disabled. If the ESC zeroes them immediately, the only real delay is `BLADEMOTOR_REVERSE_OFF_MS = 1000`, well below a blade's free-wheel time (several seconds), and the firmware would command reversal on a still-spinning blade — the exact mechanical risk this PR is meant to cover. The PR's own checklist leaves "stop-before-reverse timing" unchecked. Please run a supervised bench test, **blades removed**: OFF from full speed, trace `BLADEMOTOR_u16RPM` / `BLADEMOTOR_bActivated` for several seconds. If RPM decays progressively the guard is sound; if both drop to 0 instantly, raise `BLADEMOTOR_REVERSE_OFF_MS` to measured free-wheel time + margin and document it.

**Medium (small, please fix in the same pass)**
- `blademotor.c:302` — the early `break` on `gState != HAL_UART_STATE_READY` now also skips the ESC error check (`pu8ReceivedData[6] != 0`) and the `HAL_UART_Receive_DMA()` re-arm that `dev` ran unconditionally every 100 ms. Move the guard just before `blademotor_prepareMsg()` and keep the error check above it.
- `blademotor.c:96-131` — if a controller never produces a qualifying sample (non-zero RPM noise, `activated` stuck, no feedback), the reversal never completes and the blade silently never starts. Add a throttled `debug_printf` or a `BLADEMOTOR_u32Error` increment when `blademotor_reverse_pending` exceeds N seconds so this is diagnosable in the field (with #558 the UI would otherwise show "Reverse" while nothing turns).

**Low**: CI job name no longer describes its content; `fetch-depth: 0` is an out-of-scope fix; `BLADE-REVERSE.md` not indexed in `docs/claude/doc-index.md`.

Merge order: this PR before #558.

wjcloudy:
@cedbossneo I have pushed the software fixes and measurement preparation in **0c03981f**. **The HIGH coast-down finding remains open.** I do not have a capture proving that ESC RPM tracks the rotor after OFF, and have not substituted an arbitrary longer delay.

- **TX busy:** ESC error handling and receive DMA re-arming now precede the TX-buffer ownership guard. The regression checks that a busy UART still consumes the error and cancels ON, re-arms RX, and leaves the DMA-owned frame unchanged; clearing the error cannot revive that cancelled request.
- **Pending reversal diagnostics:** after 5 seconds, increment `BLADEMOTOR_u32Error` and emit a throttled debug message every 5 seconds. Tests cover nonzero noise, active/invalid/missing feedback, busy TX, repeated ON, cancellation and tick wrap. No warning timeout permits reversal.
- **Low-priority items:** renamed the combined board/blade CI job, removed the unrelated build-matrix `fetch-depth: 0` change, and indexed `BLADE-REVERSE.md`.
- **Physical blocker preparation:** the operator has a standard 500 on the bench. I built `Yardforce500_COASTDOWN_VALIDATION` (non-LFP): normal gated forward/OFF, no auto-start, and reversal held OFF indefinitely even with qualifying feedback. Foreground UART logs include sample sequence/receive time, validity, RPM/activated/error, and the last accepted command-transition time. The native harness checks that mode too. The document supplies the blades-removed procedure, exact-baseline requirements and pass/fail/inconclusive criteria against an independent optical tachometer. Sequence gaps must be treated as missing evidence. A blades-removed run cannot establish a loaded blade's worst-case coast time.

Both standard firmware builds and the validation build pass at **1.9.206**; production blade tests pass for both board selections and bench mode, and all three existing Unity safety tests pass under Linux. Board/default, protocol, generated-header and diff checks pass. CI has been triggered separately.

The 500 measurement is **HARDWARE_PENDING**, and equivalent 500B evidence is **HARDWARE_REQUIRED**. The previous direction reports remain scoped to their earlier exact baselines. No firmware was flashed or motor operated for this follow-up. I will bring back the coast-down capture before asking you to clear that blocker; merge order remains #559 before #558.

cedbossneo (commented):
Re-checked 0c03981f: the TX-busy guard now sits after the ESC error check and RX DMA re-arm, and the pending-reversal diagnostic (BLADEMOTOR_u32Error + throttled debug after 5 s) is in with tests; nothing else changed; no conflict with dev. The remaining item is the bench measurement, as you noted: blades removed, OFF from full speed, trace `BLADEMOTOR_u16RPM` and `BLADEMOTOR_bActivated` for a few seconds. If RPM decays progressively the 1000 ms interlock stands; if both drop to zero instantly, raise `BLADEMOTOR_REVERSE_OFF_MS` to the measured coast-down plus margin. Post the capture here and I'll merge (then #558 right after).

cedbossneo:
Merged current `dev` into your branch (d1e47b9a), clean, no conflicts — just to keep CI on a fresh base. The open item is unchanged: the coast-down capture showing ESC RPM tracks the rotor after OFF.

wjcloudy:
@cedbossneo Testing on the original 500 (`962ce210` with validation instrumentation) showed that the ESC **holds its last nonzero speed report after OFF, then jumps to zero**, rather than reporting progressive spin-down. Polling continued at ~100 ms with valid replies: `active` cleared promptly, the speed word held at `3494`, then cleared at ~2.26 s. I observed the rotor stopped before the reported zero. The additional response words showed no progressive coast signal in this capture.

I've pushed **c942ba31** to reflect this:
- Identify the value as a reported speed word in the driver and validation trace; distinguish reply freshness from a new physical speed measurement.
- Require a **new qualifying reply to release reversal**. A cached zero can no longer release it merely because the minimum OFF dwell expires.
- Add regressions for the hold-then-zero pattern in both directions, longer holds, interrupted zero confirmation, continuous OFF polling and cached-zero rejection. The full 300 ms confirmation starts at the first qualifying zero; there is no forced-start timeout.

All three firmware builds and regression configurations pass, alongside protocol/default checks. [Testing details and baseline](https://github.com/wjcloudy/mowglinext/blob/c942ba31/firmware/stm32/ros_usbnode/BLADE-REVERSE.md#recorded-standard-500-result-2026-09-19).

This corrects my earlier overstatement: the capture does not distinguish an ESC timeout from a physical stop detector. Broader stop validation and equivalent 500B evidence remain outstanding. Please review the fix and these testing limits.

wjcloudy:
@cedbossneo Tested the updated standard 500 firmware (`c942ba31`, **1.10.233**) with cutting blades removed. I visually confirmed **forward → complete stop → reverse → complete stop**, including the short paired comparison. USB telemetry captured spin-up and return to zero on each run.

This confirms explicit direction changes with a stop between requests on the original 500. It does not establish automatic reversal while coasting or equivalent 500B behavior. [Test record](https://github.com/wjcloudy/mowglinext/blob/ac5d167b/firmware/stm32/ros_usbnode/BLADE-REVERSE.md#standard-firmware-direction-test-2026-09-19).

wjcloudy:
@cedbossneo I've added both sensor recovery routines in **0f7d8531 / deefc5d2** and updated the PR description:

- **External IMU:** bring the existing bounded software-I2C bus-clear routine upstream, with limited retries and invalid reads preserved.
- **Onboard tilt sensor:** recover the hardware-I2C bus in stages in the main loop, then verify identity and restore/read back its safety configuration. USB callbacks now read a fresh, valid snapshot instead of doing blocking sensor transactions.
- **Safety:** an unavailable onboard sensor inhibits motion. A detected fault stays latched after recovery and requires explicit release with healthy, untriggered sensor status; a heartbeat reconnect cannot clear it.

The production-code fault-injection tests cover both board variants, stuck lines, failed transactions, lost configuration, stale status and the real heartbeat callback. Firmware CI is green at **deefc5d2**, including the standard 500/500B builds and blade/recovery regressions. The new automatic recovery itself remains **HARDWARE_PENDING**; the earlier manual bus-clear result does not validate this implementation. [Recovery design and acceptance procedure](https://github.com/wjcloudy/mowglinext/blob/deefc5d2/firmware/stm32/ros_usbnode/I2C-RECOVERY.md).

The earlier blade review fixes remain in place. The standard 500 test confirmed forward → complete stop → reverse → complete stop, but the ESC holds its reported speed after OFF and then clears it: it does not expose progressive coast-down RPM. That leaves the stop-detection interpretation and equivalent 500B coast-down acceptance outstanding; I am not marking your physical blocker closed.

Please re-review the recovery addition and the recorded testing limits. Once you accept the remaining evidence/acceptance points, could you merge **#559 first, then #558** alongside it?

wjcloudy:
@cedbossneo Flashed and tested the recovery firmware on both **500 and 500B**: flash verification, USB reconnection, sensor telemetry and charging checks passed, without a physical power cycle.

Tested builds: 500 **1.10.236** (`deefc5d2`); 500B **1.11.17** (`e6ced900`). No motors were run during these deployment checks.

Neither board encountered a bus fault, so automatic recovery from a real fault remains unverified. The previously recorded coast-down test results are unchanged.

PR #564, opened 2026-09-06
feat(coverage): fix Auto angle and add per-area cross-hatch controls

wjcloudy:
## What this adds

Add **Cross-hatch mowing** and per-area next-direction controls to Settings → Mowing. Each lawn can alternate between its normal stripe angle and that angle +90°, helping vary the mowing direction between sessions. Operators can also choose the next direction explicitly without changing an active or paused run.

This also fixes large-field Auto angle selection: a negative longest-edge heading previously looked like the Auto sentinel and unexpectedly triggered the exhaustive search.

## How it works

- Cross-hatch defaults to **off**. Enable it with the existing save/restart flow. Fixed 25° alternates 25° / 115°; Auto alternates Auto / Auto +90°, with the actual base angle resolved per polygon.
- Each area owns its phase. Attempting coverage consumes at most one phase at session end (recorded before the blade-service availability check, not confirmation of physical rotation); selected-area mowing only advances that area. Pauses, recharge, replanning and host restarts retain the current direction. Failed planning does not silently consume a phase or select another direction.
- **Next stripe direction by area** shows current/next direction. Overrides save immediately, apply to the next session and survive restart. The panel uses the running configuration, not unsaved settings.
- State shares the existing coverage resume file. Session end removes the old command/cursors before retaining phase metadata. Never-enabled users keep no completed cross-hatch history. Storage failures are reported; failed explicit overrides roll back.
- Orientation requests are queued for the behaviour-tree owner, validated against the ROS map, and acknowledged only after saving. Invalid/navigation-only areas and unavailable validation are rejected. Persistence remains synchronous on the BT owner, consistent with the existing resume writers.
- Original ROS area IDs survive the GUI's filtered list and editable mowing order. Direction controls, area-select mowing and obstacle promotion target the correct lawn even with interleaved navigation areas. Unknown IDs disable the action.
- Repeated unsuccessful planning produces area-specific advice and an additional warning after three failed sessions. Coverage start or an explicit override clears the streak.

## Planning and compatibility

Perpendicular generation uses the resolved fixed/large-Auto angle directly; exhaustive Auto uses the longest valid swath. Empty, single-point or coincident clips cannot silently select absolute 90°.

The boundary, headland rings, obstacle constraints, minimum radius and on-edge tolerance remain unchanged. Rotated rows meet different edges, so both orientations run through the full continuous-path envelope tests. Before retaining a straight connector above the existing 120° sharp-turn limit, the planner tries other bounded Dubins candidates. This connector fallback applies to ordinary mowing too, independently of the cross-hatch setting. Ordinary accepted pivot joins retain their behaviour. Curved alternatives can add travel distance; the recorded-area fixture produces one subpath in each orientation, with driven lengths of 191.41 m / 261.85 m. Geometry checks do not establish field tracking performance.

Rebuild/upgrade the behaviour tree, coverage server and GUI together: `PlanCoverage.action` adds `perpendicular`, and `CoverageOrientation.srv` exposes read/set-next through the GUI proxy. Firmware and its protocol are unchanged. This is independent of blade motor direction.

Phase history follows ROS area indices, as existing coverage resume state does. Review next directions after deleting/reordering areas in the saved map. [Operator details and commissioning procedure](https://github.com/wjcloudy/mowglinext/blob/codex/coverage-crosshatch/docs/CROSS_HATCH_MOWING.md).

## Screenshots

Actual GUI rendered in Chrome with **mocked robot data**, refreshed for this revision.

Auto: each lawn has its own current and next direction.

Auto cross-hatch settings

Changing Back lawn's next direction leaves its active run and Front lawn unchanged.

Auto next-direction override

Fixed 25°: the next perpendicular session is 115°.

Fixed-angle cross-hatch settings

The next fixed-angle direction can also be overridden independently.

Fixed next-direction override

## Validation

- Connector follow-up (`8d8bfe74`): **59 planner tests pass**, including four dedicated alternate-Dubins regressions through the public continuous-path planner. They cover nominal and reduced-radius curves, boundary rejection and hole rejection. Disabling only the second Dubins search in a temporary copy makes both positive cases fail; the two rejection cases still pass. clang-format 18 changed-line and diff checks pass. This follow-up ran as UID 1000 against Fields2Cover 3 in the cached Kilted test image `sha256:826e6868aaa794a4a03e5585259aa0ffee88a21e65557ba148c01389f3e59218`; the earlier validation below retains its original dependency baseline.
- Non-root ROS2 Kilted builds with Fields2Cover `884d895b59192882476e986ba44ea9143a06a6a9`; ROS results: **464 reported checks, 0 errors, 0 failures, 61 skipped**.
- Original validation: **55 planner tests**, including both orientations of the recorded boundary, swept footprint, zero-slack, rings-off and fragmentation fixtures; fixed-angle +90° equivalence tested through the full driven path.
- Behaviour/service/persistence regressions cover unavailable blade control, independent phases, toggle/restart, invalid map IDs, deferred ordering, bounded multi-threaded request handling, service destruction, failed writes and stale-START protection.
- Go backend/map/schema tests, **393 GUI unit tests**, TypeScript and ESLint passed. One lint-config unit-test timeout under parallel load passed on its isolated rerun; lint warnings remain.
- Four Chrome browser cases passed: fixed/Auto overrides, reload and unavailable ROS, plus actual obstacle-promotion requests after mowing-order changes and missing-ID rejection. Screenshots were regenerated and inspected.
- Ten launch-injection tests, config drift, generated Go/TS/firmware bindings, XML validation and clang-format 18 changed-line checks passed. The template count is verified from YAML by `refresh_parameter_summary.py`.

**HARDWARE_REQUIRED:** the revised physical routes still require supervised commissioning. The operator document records the software baseline, required deployment identities, procedure and pass/fail criteria. No robot was deployed or operated for this review update.


cedbossneo (changes requested):
Careful, well-documented PR — the state machine is small and readable, the persistence ordering is genuinely well reasoned, and the Auto-angle fix is a real bug fix. One blocking gap, which is about test coverage rather than about the code being wrong.

## What I verified as correct

- **Invariant 15 satisfied.** `mow_cross_hatch: false` lives only in the in-package template; the GUI schema default matches; the installed sparse config is untouched; launch injection is guarded by a new test.
- **The `working_area_indices` change is a genuine fix, not a regression.** `map_server`'s `areas_` is a flat list including navigation areas (`area_manager.cpp:571`), so the GUI's `working_area` ordinal was never the ROS index. Selecting "mow area N" with an interleaved navigation area could target the wrong area. Good catch, and it needed fixing.
- **`longestEdgeAngle` normalisation is right.** `atan2` returns `[-π, π]` and negative *is* the AUTO sentinel, so an unnormalised edge heading silently re-entered the exhaustive search. Normalising mod π fixes it, and the new regression test pins both windings and both rotations.
- **`clearCoverageResumeState` ordering is correct and the comment earns its place.** Removing the file *before* writing the phase-only snapshot means a crash in between loses phase history but can never resurrect a stale `COMMAND_START`. That is the right trade and it is the safe direction to fail in. `has_resumable_state` correctly keeps a phase-only snapshot from auto-starting on boot.
- **Rings are genuinely untouched** by the rotation — they are concentric boundary offsets, not angle-dependent.

---

## [BLOCKING] The rotated layout never reaches a single continuous-path or in-bounds test

`test_coverage_planning.cpp:2350-2375`

`perpendicular=true` appears at exactly two places in the whole suite — lines 2360 and 2361 — and **neither result is passed to `buildContinuousSubPaths`**. Every one of the eight `buildContinuousSubPaths` call sites (`:1286, :1287, :1360, :1428, :1701, :2173, :2314, :2317`) runs a plan built with the default `perpendicular=false`.

The one new geometry test asserts swath **endpoints** against an axis-aligned rectangle. That is not the geometry the robot drives. What it drives is rings + swaths joined by the forward turn-around arcs that `buildContinuousSubPaths` builds *between* swaths, bounded to `connector_clearance_boundary` — and clipping swaths to `mainland` bounds the swath, not the connector arc constructed afterwards.

So none of these guards ever sees a rotated plan:

- `TurnArcFootprintStaysInsideRecordedBoundary` (`:1223`)
- `RecordedArea1NoCuspInBounds` (`:1671`) — the real recorded field
- `OutermostRingStaysInsideClearanceRingOnTheLine` (`:1336`)
- `HoleFreeFieldIsOneSubPath`, `NotchFieldLobeChainedNoMidFieldJoins`, and the rest of the fragmentation set

**Why this specifically matters here.** Per Invariant 7 the connector clearance ring is `safe_boundary` **exactly, with no outward slack** — an outward expansion was rejected precisely because it would let a turn centreline cross the recorded line — and on-edge connectors are accepted only within the 1 mm `kOnEdgeTolM`. Rotating the serpentine 90° puts the row ends, the tightest geometry in the plan, against the *other* pair of boundary edges and against a headland apron of different local shape. Two outcomes, neither covered:

1. A U-turn connector fails `allInside` → the sub-path **fragments**, converting one continuous blade-on pass into several blade-off Nav2 transits mid-lawn — exactly the regression the fragmentation tests exist to catch.
2. A connector is accepted on-edge but its swept chassis footprint crosses the recorded line — exactly what `TurnArcFootprintStaysInsideRecordedBoundary` exists to catch, on a machine with spinning blades.

The PR declares itself *"Safety-critical: changes physical mowing routes"* and leaves `[ ] Hardware/simulation commissioning` unticked. That combination is fine — field time is scarce — but then the simulated envelope guards have to carry the weight, and right now they are not exercised at all on the new path.

**The fix is cheap:** parameterise those tests over `perpendicular ∈ {false, true}`. They already call both `planBoustrophedon` and `buildContinuousSubPaths`; only the last argument changes. `RecordedArea1NoCuspInBounds` and `TurnArcFootprintStaysInsideRecordedBoundary` are the two that matter most — if the rotated plan passes those on the real recorded field, I am satisfied.

---

## Nits

- `coverage_planning.cpp:1216-1218` — `swaths[0].getPath()` is dereferenced with no `line.size() < 2` guard, and `line.size() - 1` underflows on an unsigned type if the path is empty. Six lines below, at `:1224`, the ordered-swath loop *does* guard exactly this (`if (line.size() < 2) continue;`), so the file already treats a degenerate generated swath as possible. Worth matching that posture.
- `coverage_planning.cpp:1218` — the regenerated `swaths` is not re-checked for emptiness before `genSortedSwaths`. I no longer think this is a real coverage hole (rotation makes a narrow cell *denser*, and anything that empties out is already instrumented via `min_swath_length` → `plan.diagnostics.drops`), but it is one line and sits right next to the item above.
- `coverage_orientation_service.hpp` — `ctx->cross_hatch[req->area_index]` inserts on any index with no bounds check, and each insertion is persisted. The GUI only sends valid indices, so this is defensive only, but it is an unvalidated write at a system boundary that grows the resume file.
- `mower_config.schema.json` — the new `mow_cross_hatch` block is indented 4 spaces against 2/10 for its neighbours, and unlike them has no `description`.
- **Adjacent, pre-existing, not yours to fix here but now a one-liner:** `MapPage.tsx:393-406` (`obstacleAreaIndex`) still uses the filtered workarea ordinal. Its comment justifies this by "workareas are written first, then navigation areas" — true for a GUI-saved map, false for a robot-recorded one with an interleaved navigation area, where `promote_obstacle` will stamp the NO_GO_ZONE into the wrong area. Now that `mowingAreaIndex()` exists, a follow-up PR could close it.

---

## Verdict

**Request changes** — on the test gap alone. The implementation looks right to me; I just will not sign off on a rotated mowing route whose connector geometry has never been through the boundary guards this repo built for exactly that question. Add `perpendicular=true` to the two envelope tests and I will re-review promptly.

Heads-up: this PR now conflicts with `dev` on `gui/pkg/msgs/mowgli/types.go` and `gui/pkg/providers/ros.go` — #574 landed and turned the dock fields into optional pointers. Both conflicts are mechanical.

cedbossneo:
Follow-up to my review — a second pass sharpened the `swaths[0]` nit into something worse than I first wrote it, and turned up four more. Grouping them so you can do one pass.

### The `swaths[0]` sampling has a second, quieter failure mode (upgrading this from nit to MEDIUM)

`coverage_planning.cpp:1211-1218`

I flagged the unguarded `line.size() - 1` underflow. That is the *loud* half. The quiet half is worse:

If `line.size() == 1`, or the two sampled endpoints coincide, then `a == b` → `atan2(0, 0) == 0` → `angle = fmod(0 + π/2, π) = π/2`. `bf.generateSwaths(π/2, op_width, cell)` then mows that entire cell at **90° absolute in the map frame** — not `base + 90°`. The operator's Auto selection or fixed `mow_angle_deg` is silently discarded for that cell. No drop note, no diagnostic, and `plan.swath_angle_rad` at `:1239` reports a wrong-but-entirely-plausible value, so nothing downstream can tell.

A crash is loud and gets reported. A silently wrong stripe direction on someone's lawn is not.

What makes this more than theoretical: `swaths[0]` is the **unsorted** first swath — the extreme of the sweep, structurally the most likely to be a tangent or degenerate clip, especially against the rounded convex corners `generateHeadlands` produces. Sampling the single worst candidate is the design flaw; the missing guard is only how it surfaces.

**Suggested fix:** stop sampling a swath at all. Use `cell_angle` directly when it is `>= 0`, and for the AUTO branch derive the heading from the **longest** swath rather than the first. Re-checking emptiness after `:1218` then falls out of the same edit — which also closes the `size() == 0` point I raised and have since dropped as a standalone concern (you were right that rotation makes a narrow cell denser, not emptier, and that `min_swath_length` already instruments the drop via `plan.diagnostics.drops`).

### `coverage_nodes.cpp:1125-1135` — the `used` latch sits behind a 200 ms service wait

`used` is the sole gate that consumes a phase, and it is latched *after* the `wait_for_service(200ms)` early return. One transient `mower_control` miss and that area never sets `used`, so `finish()` never flips it and **that area silently stops alternating forever** — while every other area keeps cross-hatching. Latch the phase consumption before the service call, or on a signal that does not depend on the service being up.

### `cross_hatch.hpp:57-66` + `coverage_server.cpp:514` — a repeatably-failing perpendicular plan wedges the area

If the perpendicular plan fails repeatedly for an area, `used` is never set, the phase never flips, and the area is skipped every session thereafter with no diagnostic. `finish()`'s "a failed plan does not consume an orientation" rule is right for a one-off failure and wrong for a persistent one. Worth at least a warning when an area's phase has been pending across N sessions.

### `coverage_nodes.cpp:1129,1853` — unconditional `operator[]` makes every user pay for the feature

`ctx->cross_hatch[area_index]` inserts an entry whether or not `mow_cross_hatch` is enabled. Since `clearCoverageResumeState` branches on `!ctx->cross_hatch.empty()`, EndSession becomes `unlink + rewrite` instead of a plain `unlink` for **all** users — including everyone who never turned the feature on — and hands them a new storage-failure ERROR path for a feature they are not using. Guard the insertion on `ctx->mow_cross_hatch`, or use `find` on the read paths.

### `coverage_orientation_service.hpp:43-49` — inline context mutation + blocking SD I/O from a callback

The service mutates `BTContext` and does blocking SD-card I/O directly in the callback, against the deferral convention documented at `behavior_tree_node.cpp:660-668` (which queues work for the next tick). It is safe today because both live in the same MutuallyExclusive group — the same implicit assumption I flagged on #558. Since neither is enforced by a test, it is worth following the existing convention rather than adding a second place that quietly depends on the group.

### Doc nits

- `CROSS_HATCH_MOWING.md:37` — the empty-path claim contradicts the rollback at `coverage_orientation_service.hpp:46-51`.
- `coverage_nodes.cpp:1133,1857` — `saveCoverageResumeState` returns swallowed; EndSession logs its failure, these do not.
- `parameters.md:36` — stale "158 template keys".

---

Priorities unchanged: **only the test parameterisation blocks.** I would like the `swaths[0]` sampling and the `used` latch in the same batch — both are short and both are in code you will be touching anyway. The rest can be follow-ups; say the word and I will not hold the merge on them.

For the blocking item, in value order: `RecordedArea1NoCuspInBounds` (`:1671`) first — the real recorded field is the highest-value one by some margin — then `TurnArcFootprintStaysInsideRecordedBoundary` (`:1223`). `OutermostRingStaysInsideClearanceRingOnTheLine` and `HeadlandDisabledIsOneSubPath` cover the zero-slack and rings-off cases if you want the full set. A two-iteration `for (bool perp : {false, true})` is plenty.

wjcloudy:
I've addressed both review comments in **92f02d06**, including the suggested follow-ups. Commits are separated into the dev merge (**91d9068b**), planner/tests (**44cc2478**), behaviour/persistence/service (**06dd0579**), GUI IDs (**25e1a668**) and documentation/screenshots (**92f02d06**).

1. **Blocking rotated-path coverage:** parameterised `RecordedArea1NoCuspInBounds`, `TurnArcFootprintStaysInsideRecordedBoundary`, `OutermostRingStaysInsideClearanceRingOnTheLine`, `HeadlandDisabledIsOneSubPath`, `HoleFreeFieldIsOneSubPath` and `NotchFieldLobeChainedNoMidFieldJoins` over both orientations. Each reaches `buildContinuousSubPaths`. Clearance, footprint, 1 mm on-edge tolerance, sharp-turn ceiling and fragmentation upper bounds are unchanged. The notch's perpendicular sweep legitimately needs one subpath, so only its orientation-dependent minimum count differs.

   The tests exposed a 135° straight fallback in the rotated recorded fixture. Before retaining a connector above the existing 120° limit, the planner now tries the other Dubins candidates within the same boundary/radius limits. Both recorded orientations pass at 91.05° maximum local turn, one subpath and zero out-of-bounds sampled points. **Trade-off:** curved alternatives add travel; current fixture lengths are 191.41 m base / 261.85 m perpendicular. Ordinary accepted pivot joins and the existing apron/fallback-statistics regressions are preserved. This deserves review alongside the test additions.

   I also added a full-path equivalence test: perpendicular fixed-angle planning equals selecting `(angle + 90°) mod 180°` manually. The distinction is test coverage, not a changed boundary definition: different row ends create different connectors against that same boundary.

2. **First-swath/degenerate-angle issue:** fixed and large-Auto paths rotate `cell_angle` directly. Exhaustive Auto selects the longest finite, non-degenerate swath, with empty/single-point/coincident clip regressions. Missing usable direction is diagnosed, and rotated output is checked before sorting.

3. **`used` latch:** coverage-start bookkeeping now happens before the 200 ms blade-service wait and is idempotent. A real `FollowStrip` test covers unavailable blade control, repeated startup, halt and repeated session end. This records a coverage attempt; it is not evidence of physical blade rotation.

4. **Repeated failed orientation:** retain the phase, log area/orientation and override advice, and warn again after three distinct failed sessions. Retries don't inflate the count. The streak persists with phase metadata and resets on coverage start or explicit override; there is no automatic angle fallback or phase consumption on failure.

5. **Disabled-feature writes:** removed unconditional history insertion. Never-enabled sessions end with unlink only. Temporary base-direction records accompany normal resume progress so enabling cross-hatch after restart cannot rotate an interrupted disabled run. Legitimate prior history remains intact. Toggle/restart/progress-reset regressions cover this distinction.

6. **Service validation and ownership:** callbacks enqueue bounded requests; BT-owned processing performs asynchronous map validation, applies the edit and saves before sending a deferred success response. Invalid indices, navigation-only areas, `UINT32_MAX`, unavailable validation, failed persistence/rollback, ordered edits around session end, queue saturation with a multi-threaded executor and deferred-client destruction are covered. SD writes remain synchronous on the BT owner; this follows the existing deferral convention without introducing an independent writer that could resurrect old resume state.

7. **Persistence/docs nits:** first-plan and coverage-start save failures now warn with area/path. The document distinguishes automatic in-memory phases from explicit overrides, which fail and roll back with an empty persistence path. It also explains remove-before-phase-write ordering. The current template actually contains **158 keys**; I verified it from YAML and added a reproducible refresh/check script rather than guessing a replacement count.

8. **Schema, adjacent ID bug and dev conflicts:** fixed indentation and added help. Obstacle promotion and its confirmation label use original ROS IDs, independently of mowing order; the browser test verifies the actual request after reordering lawns with an interleaved navigation area. Missing metadata disables promotion. The dev merge retains #574's optional dock fields alongside `working_area_indices`; generated bindings and Go map/dock tests pass.

Validation: ROS reports **0 errors / 0 failures** (61 skipped), all **55 planner tests** pass, and Go, GUI unit/type/lint, four Chrome browser cases, ten launch-injection checks, configuration/codegen checks and clang-format 18 checks pass. PR description and visible screenshots are refreshed. **Field commissioning remains HARDWARE_REQUIRED**, with the baseline/procedure documented; no robot was operated for this update.

@cedbossneo could you please re-review? I have addressed all the points above; the connector correction and its travel-distance trade-off particularly deserve another look. Current-head ROS, Go, GUI unit/type checks, formatting and static analysis pass, as does the arm64 image build. The amd64 image build failed only while downloading the PlatformIO installer: curl reported an SSL connection reset from raw.githubusercontent.com. I tried rerunning it, but GitHub requires upstream admin rights; could you also rerun that failed job? The documented field commissioning remains outstanding.

cedbossneo (commented):
Follow-up review:

- The blocking test gap is closed: `INSTANTIATE_TEST_SUITE_P` at `ros2/src/mowgli_coverage/test/test_coverage_planning.cpp:2386` (commit `44cc2478`) covers the `perpendicular=true` envelope. Thanks.
- The failing `Build (linux/amd64)` check is a PlatformIO TLS flake, not this PR; re-run triggered.
- **One thing that still needs a real look before approval:** the same commit `44cc2478` also changes `buildConnector` in `ros2/src/mowgli_coverage/src/coverage_planning.cpp:565-598` (plus `sampleDubins` `:430-464`): a second Dubins search loop for near-U-turn joins, gated by `kMaxPivotTurnCos = -0.5`, **not** conditioned on `perpendicular`. It therefore changes default connector geometry for every deployed robot, not just cross-hatch users. On the merits it looks like an improvement (verified `allInside` + `clearOfHoles` curves instead of blind straight joins), but it is a physical-behaviour change that arrived after the review and has no dedicated test. Please add a test for the new Dubins fallback (or split that hunk into its own PR) — then this is good to go.
- Minor: `markCoverageStarted` consumes the cross-hatch phase on a blade *request*, not a confirmed blade start.

Note for the merge queue: `feat/lidar-map-anchor` (#565) also touches `buildConnector` (`40d0c30b`, straight-fallback tangency check); whichever lands second will need a careful merge of that function.

wjcloudy:
@cedbossneo I have addressed your [follow-up review](https://github.com/mowglinext/mowglinext/pull/564#pullrequestreview-5186487023) in **8d8bfe74**.

You are right that the alternate Dubins search affects ordinary mowing too. I have added four dedicated `CoverageAlternateConnector` regressions in `test_coverage_planning.cpp`, using fixed ordinary swaths through **`buildContinuousSubPaths`**, without a perpendicular/cross-hatch option:

- **Longer in-bounds word:** a 135° straight departure where the shortest RSL leaves the boundary but the longer LSL fits. Checks arc classification, endpoints, route length, smooth joining, sampled containment and the turning-radius floor.
- **Radius reduction:** a raised lower boundary rejects the nominal 0.18 m curve but permits the 0.16 m curve, above the 0.15 m floor.
- **Boundary rejection:** no alternate curve fits; the clear straight join remains classified as a fallback.
- **Hole rejection:** an obstacle blocks the otherwise fitting curves without blocking the swaths/direct join; those curves must be rejected too.

**Regression proof:** I disabled only the second search in an isolated temporary source copy. Both positive tests fail, reverting to straight joins with a 135° turn; the two rejection tests still pass. With the actual PR code, **all 59 planner tests pass**, including both orientation suites. clang-format 18 and diff checks pass. The PR validation section records the exact local test-image baseline; upstream CI is running separately for this push.

This commit changes tests and documentation only. I also clarified your minor point: phase consumption records a coverage attempt before checking blade-service availability, **not confirmed blade start/rotation**. The shared `buildConnector` merge consideration with #565 still applies; this fix adds no further edits to that function. Hardware commissioning remains `HARDWARE_REQUIRED`, with the existing procedure documented; no robot was operated.

Could you please take another look?

cedbossneo (commented):
Thanks — the four `CoverageAlternateConnector` regressions are real (I re-derived the Dubins geometry independently: without the second search the two positive tests fall back to `straight_kept`, with it LSL at r=0.18 / 0.16 wins), and the hunk itself is sound (radius never below `min_radius`, `allInside` + `clearOfHoles` on every candidate, shortest valid word wins, no cusp).

What blocks now is the rebase: `dev` moved under the PR with #565 (merged this morning) and it is not mechanical. `git merge-tree origin/dev pr-564` → 4 conflicts:
- `ros2/src/mowgli_coverage/include/mowgli_coverage/coverage_planning.hpp`: dev changed `planBoustrophedon(..., double min_turn_radius = 0.20)`; the PR keeps `0.15` and adds `perpendicular = false`. Please keep **0.20** (deliberate physical change: `mowgli_robot.yaml min_turning_radius: 0.20`, `num_headland_passes: 5`, `navigation.launch.py`).
- `ros2/src/mowgli_coverage/test/test_coverage_planning.cpp`: dev added `straightFallbackIsContinuous` (40d0c30b) wired into `conn_safe` — a straight fallback that is not tangent within 15° is now a **split**, not `straight_kept`. Your two negative tests (`BoundaryRejectsAllAlternateWords`, `HoleRejectsOtherwiseInBoundsAlternateWords`, 135° join) will therefore see `split=1, paths=2` after the merge; their expectation should become "split, no arc" rather than "straight_kept".
- `docs/claude/parameters.md`, `gui/pkg/api/mowglinext_test.go`: textual.

After that rebase (and CI green) this is good to go.

cedbossneo (approved):

PR #565, opened 2026-09-06
feat(fusion_graph): LiDAR map anchor — scan-to-map localisation for RTK-Float (Beluga)

cedbossneo:
## What

A **scan-to-map LiDAR anchor** for RTK-Float, replacing the scan-to-keyframe ICP anchor (which is removed in this PR).

- Under fresh RTK-Fixed every scan is ray-cast into a georeferenced log-odds grid (`/fusion_graph/lidar_map`, 800×800 @ 0.10 m, latched; the GUI map page draws it via #566).
- Once Fixed goes stale (> 1.0 s, with a 1.0 s disengage dwell — 0.3 s flapped on a 5 Hz receiver whose Fixed age idles at 0.08–0.28 s) a **Beluga** particle filter (`beluga_ros::Amcl`, likelihood field, differential-drive motion from the node's own dead reckoning) localises against that grid.
- Each estimate becomes an XY-only `PoseTranslationPrior` **only if it passes three independent gates** (`lidar_anchor_validator.hpp`): scan consistency at the estimate (≥ 50 % of the full scan's returns on mapped occupied cells, ≥ 30 of them), particle spread ≤ 0.5 m, and plausibility vs dead reckoning (0.3 m + 2 % of the path since seed). After 5 s of rejects the cloud is re-seeded from dead reckoning.
- **Shadow mode** (`lidar_anchor_shadow_mode`): run, score and publish the filter under RTK-Fixed too, never apply — the field measurement of anchor vs RTK (`/fusion_graph/lidar_anchor_candidate`, diagnostics `lidar_anchor_*`).
- **Cleanup**: the keyframe anchor (`graph_manager_keyframe.cpp`, `kf_*` params, `.keyframes` persistence, `kf_matches_*` diagnostics, the GUI "ICP Keyframes" tile) is deleted, −1159 lines; unique coverage it held (datum guard, `RigidTransformAll`) is ported, and the live anchor factor gets the equivalent of the deleted `KeyframeApply` tests.
- **Replay harness** (`ros2/src/fusion_graph/tools/replay/`): bag → node → scored against dead reckoning and GPS, on a laptop, in 10 minutes. It reproduced the field failure and proved the fix.

## Field + replay evidence

Field 2026-09-06 (GPS container stopped 2 min 40 s, operator driving slowly): the first version put the pose **10 m** off; wheels + gyro alone were within 0.4 m. Root cause found by replay: `beluga_ros::Amcl`'s motion model reads a 2-pose odometry window with no first-update special case, so the first update after construction applies the *whole* odom pose (≈ 3 m) instead of a delta — the cloud teleports into the terrace clutter and never returns. Fixed with one discarded warm-up update after construction; the gates remain as defence in depth.

Replay on the 2026-09-06 bag, robot's own map imported (max ‖fused − dead reckoning‖ during the 158 s outage / fused error at RTK return):

| Configuration | max | at RTK return |
|---|---|---|
| as first deployed (no gates), 2 runs | 3.0 / 5.5 m | 2.5 / 4.6 m |
| gates only, 3 runs | 0.14 / 0.31 / 0.14 m | 0.05 / 0.12 / 0.04 m |
| warm-up only, no gates | 0.06 m | 0.13 m |
| **gates + warm-up (this PR)** | 0.12 m | 0.12 m |
| anchor off (reference) | 0.17 m | 0.03 m |

Shadow mode under RTK: anchor-vs-fused p50 0.09 m, p90 0.30 m, max 0.63 m — a fallback, not a centimetre reference.

## Defaults and flags

`use_lidar_map_anchor: false` and `lidar_anchor_shadow_mode: false` in the template (Invariant 15), both LiDAR-gated in `navigation.launch.py` (pinned by `test_lidar_map_anchor_flag_is_plumbed_end_to_end`). AMCL random injection off (`alpha_slow/fast = 0`), selective resampling on. `lidar_map_import_topic` (default off) seeds the mapper from a previously published grid — the replay's key, and the seam for map persistence.

## Test plan

- [x] 27 gtest suites green in the branch image (mapper, gate, validator, XY prior, persistence, rigid transform, …); `test_nav2_params.py` plumbing test
- [x] Static dock test: gate no longer flaps (seeds stay 0 under Fixed), map builds (dock, terrace walls)
- [x] Replay table above
- [x] Deployed on the robot with anchor + shadow mode (revision `3973309a`): hit ratio 0.955 at the dock, 0 factors applied
- [ ] Field: mow under RTK with shadow mode, compare candidate vs fused; then a controlled GPS cut on the lawn with the gates live
- [ ] Field: a real RTK-Float episode under the terrace

Depends on nothing; #566 (GUI map layer) reads the topic this PR publishes.


cedbossneo:
## Field test 2026-09-06 — DO NOT MERGE as is

Static dock test was fine (gate hysteresis holds, map builds). The **moving no-GPS test failed badly**: GPS container stopped for 2 min 40 s while the operator drove slowly, did a U-turn and stayed in a roughly known area.

At RTK return (bag `bag_lidar_rtk_20260906_2128`, GPS projected about the datum, lever arm not removed):

| Estimate | Position error vs GPS | Yaw vs dead reckoning |
|---|---|---|
| fused (with LiDAR anchor, 168 XY factors) | **10.0 m** | 88° off |
| dead reckoning only (map→odom frozen at the outage, wheels + gyro) | **0.39 m** (includes ~0.2 m lever arm) | reference |

Scan-vs-map consistency during the outage: 13 % of beams on known walls (97 % under RTK before). The particle filter was lost (robot left the 1830-cell map into open lawn), yet its estimate went into the graph as `PoseTranslationPrior` at the 5 cm σ floor — and through the non-holonomic wheel factors the wrong XY track also rotated the yaw. The graph then took ~100 s of map→odom slew to recover once RTK was back (8.3 m / +105° correction).

Required before this can ship:
1. **Health gate** on every filter estimate: fraction of beams landing on mapped occupied cells at the estimate (≥ 0.5) AND particle spread below a cap; otherwise no factor.
2. **Plausibility vs dead reckoning**: reject an anchor estimate farther from the DR pose than DR's drift budget (e.g. 0.3 m + 2 % of distance since RTK loss).
3. σ from the filter covariance without the 5 cm floor as a *minimum of trust* — use it as a floor only when the health gate passes; cap σ (e.g. 0.5 m) otherwise skip.
4. Do not anchor when the current scan sees < N mapped cells (outside the map); dead reckoning is better there.

Positive result: under RTK Fixed the fused pose and the LiDAR map agree to 5 cm (p50) / 7 cm (p90) over 6.4 m of driving — the map itself is good.

cedbossneo:
## Replay harness: failure reproduced, root cause found, fixed

Built an offline replay of the 2026-09-06 bag (`ros2 bag play` + `fusion_graph_node` in the branch image on a Mac, seeded with the robot's own fused pose and — decisive — the robot's own recorded map, `lidar_map_import_topic:=/fusion_graph/lidar_map`). Metric: max |fused − dead-reckoning| during the 158 s GPS outage, and fused error vs GPS at RTK return.

| Configuration (robot map imported) | max ‖fused − DR‖ | error at RTK return | recovery |
|---|---|---|---|
| as deployed 2026-09-06 (no gates), run 1 | 3.03 m | 2.48 m | 64 s |
| as deployed, run 2 | 5.53 m | 4.59 m | 73 s |
| no gates, random injection off | 4.03 m | 1.88 m | 66 s |
| **gates** (validator), runs 1–3 | 0.14 / 0.31 / 0.14 m | 0.05 / 0.12 / 0.04 m | immediate |
| gates + injection off | 0.13 m | 0.03 m | immediate |
| **no gates + warm-up** | 0.06 m | 0.13 m | immediate |
| **gates + warm-up** (final) | 0.12 m | 0.12 m | immediate |

**Root cause** (`3973309a`): `beluga_ros::Amcl` feeds its motion model from a 2-pose rolling window of odometry with no special case for the first update — the "previous" pose is the identity, so the first update applied the *whole* odom pose (≈3 m after minutes of driving) to every particle instead of a delta. First candidate after a 0.10 m seed landed 3.2 m away in the terrace clutter (scan consistency 0.03–0.12), then never recovered; with the 5 cm σ floor and no gate it dragged the graph. Never happens after a re-seed (window full). Fix: one discarded update right after construction. With the warm-up alone the ungated anchor already holds 0.06 m.

**Defence in depth stays** (`28fd681c`): per-estimate validator (scan consistency ≥ 0.5 at the estimate, spread ≤ 0.5 m, within 0.3 m + 2 %·path of dead reckoning), DR re-seed after 5 s of rejects, and **shadow mode** to measure the anchor against RTK in the field without applying it. Random injection off (`4ce083de`) was a wrong first hypothesis but is still the right setting (global relocalisation is never wanted here). Map import (`ac203dc3`) is the seam for map persistence.

Shadow mode under RTK on the bag: anchor-vs-fused p50 0.09 m, p90 0.30 m, max 0.63 m.

Next: deploy with `lidar_anchor_shadow_mode: true` and compare on the lawn; then a controlled GPS cut with the gates live.

cedbossneo:
## Field 2026-09-07 06:44 (shadow mode) → two more root causes, both fixed and replay-proven

**Field:** mow with `use_lidar_map_anchor` + `lidar_anchor_shadow_mode` on `a17a19ae`. First 100 s: anchor vs RTK p50 0.04 m / p90 0.09 m. Then the shadow filter teleported twice (5–8 m, tight cloud); most estimates were rejected by the gates but **6 at 5.8 m passed all three**. (The mow itself was killed by a loose LiDAR cable — `/scan` at 2.5–7 Hz, `IsScanStale` halts — unrelated; docking took 55 s once reseated.)

**Root cause 2** (`e1fad68f`): fusion_graph re-bases its raw dead reckoning every `odom_rebase_dist_m` (6 m) and resets it on set_pose / dock re-anchor. The anchor fed that raw pose to Beluga's motion model (6 m applied as a delta → teleport) and to the dead-reckoning witness (which moved by the same 6 m → the wrong estimate looked plausible). `ContinuousOdom` integrates deltas and drops any single step > 0.5 m / 60°.

**Root cause 3** (`46dee77e`): on the charger the fused pose is gauge-pinned to the dock while the BackUp undock moves the robot — scans inserted at a frozen pose, seed with the dock heading. No anchor work while charging nor 10 s after. (Also a harness bug: the seed publisher re-anchored node 0 every second; publishes once now.)

**Replay of the 06:44 bag, shadow, candidates vs RTK:**

| build | p50 | p90 | max | > 1 m | accepted & > 1 m |
|---|---|---|---|---|---|
| a17a19ae (as flown) | 0.08 | 0.22 | 7.22 | 92 | 19 |
| + ContinuousOdom | 0.08 | 0.12 | 5.00 | 58 | 17 (all in the harness-induced undock episode) |
| + charging gate, harness seed once | **0.05** | **0.09** | **0.20** | **0** | **0** |

GPS-outage bag with the final build: 0.12 m max vs DR during 158 s, 0.08 m at RTK return. Node CPU on a 9.5 min bag: 409 s → 73 s with the merged dedup (#563).

**Foliage sweep** (occupied threshold × sigma_hit × z_rand, 2 runs each on the fixed harness): no config produces a false estimate any more; precision p90 0.07–0.14 m. Two concordant hits per cell is now the default (`lidar_map_occupied_threshold: 1.7`) — same precision, and on the pre-fix harness it cut accepted false estimates 18 → 3.

cedbossneo:
## Field 2026-09-07 11:10 — first clean run, plus persistence and GUI

**Manual drive + automatic return to dock, `46dee77e`, anchor + shadow on** (369 s, 684 candidates): under RTK Fixed p50 0.04 / p90 0.08 / max 0.12 m (639 accepted, 7 rejected by dead reckoning); three Float episodes (59 s at start, two short ones near the dock on the way home): 33 accepted / 5 rejected by score, 28 factors applied at p50 0.15 / max 0.28 m. **Zero candidates beyond 0.5 m**; the 4 odometry re-bases were absorbed. Docked cleanly, pose on the dock pose, dock/GPS disagreement 0.15 m.

**Since then on the branch:**
- `cfab49f2` two concordant hits per cell by default (replay sweep: same precision, fewer false estimates through the gates on the pre-fix harness)
- `76ba61b2` the map is persisted next to the graph (`<prefix>.lidarmap`, datum-guarded, loaded at the thresholds) and `/fusion_graph_node/clear_lidar_map` empties grid + filter + file
- `cc930054` GUI: Settings → Localization toggles for the anchor and its shadow mode; Diagnostics → Fusion Graph "LiDAR anchor" tiles (state, cells, verdict + hit ratio, factors, accepted/rejected, reseeds, re-bases) and a "Clear LiDAR map" button; route tests

Still owed in the field: a long Float while driving (terrace / tree) and a controlled GPS cut with the gates live.

cedbossneo:
## Field 2026-09-08 — three mows on the branch (+ #570, #573, #576 merged in)

- **12:25 mow, 2 h 17 min, `95110115`:** started with a flaky LiDAR cable (20+ `IsScanStale` halts in 10 min) — every interrupted pass exempted (#573), **0 give-ups**, 0 LocalizationGuard pauses (#570). Then a **228 s continuous RTK-Float while mowing + returning home**: 605 anchor candidates, 396 applied, 170 rejected by the dead-reckoning gate, 20 by score, 17 by spread; graph σ stayed ~5 cm; when Fixed came back the fused pose was corrected by ≤ 0.33 m in steps ≤ 0.11 m (27.9 m driven in Float). Low-battery dock, clean docking.
- **15:15 mow (full bag, 207 MB):** 18 min until the rain sensor docked it; LiDAR intermittent again, 10 scan-stale halts all exempted; shadow p50 0.07 m under Fixed; one 16 s Float with 61 factors applied at ≤ 0.21 m.
- **Anchor-only was never actually possible until today:** the scan subscription AND the whole anchor setup block were declared inside `if (use_scan_matching_)` — fixed in `f8eef656` + `a95e1eb7`.
- **ICP vs anchor (GPS-outage bag, robot map):** DR-only 0.14 m during / **0.93 m** at RTK return (the receiver emits ~5 s of non-RTK fixes 0.9 m off before Fixed); ICP+LC 0.17 / 0.03 / 92 s CPU; anchor-only 0.13 / 0.21 / 48 s; both 0.12 / 0.05–0.12 / 92 s. ICP is what stiffens the graph at re-acquisition; the anchor is what holds a long Float. Keeping both.

PR #567, opened 2026-09-06
fix: keep mobile page content clear of bottom navigation

wjcloudy:
## Summary

At the bottom of mobile Diagnostics, **Rosbag Recording** and **ROS Diagnostics** sit underneath the fixed navigation bar and cannot be tapped. Home and Settings have the same shared scrolling problem.

The animated outlet was fixed to `height: 100%`: longer pages overflowed it, so the shell's bottom padding no longer followed their last content. Let document-style pages grow naturally, with a viewport minimum height. Keep a definite height for Map and Logs, which own their viewport layout. Include the bottom safe-area inset in the shell clearance and compensate the full-bleed map for that inset.

## Validation

- Reproduced failing clearance checks on upstream for Diagnostics, Home and Settings at 375×667; all pass with this fix.
- 11 Chromium checks passed: touch expansion/collapse of the final diagnostics panels at 390×844, 375×667 and 740×360; Home/Settings clearance and a final settings radio control; map/log viewport sizing on mobile and desktop.
- 386 frontend unit tests passed across 51 files.
- TypeScript check and production build passed. Full frontend lint passed: zero errors, 821 existing warnings. `git diff --check` passed.

## Screenshots

Mocked data at 375×667; no live mower or garden data.

| Diagnostics before — bottom panels covered | Diagnostics after — both panels reachable |
| --- | --- |
| Diagnostics before | Diagnostics after |

| Home — final health rows visible | Settings — final controls above the save/navigation bars |
| --- | --- |
| Home | Settings |


cedbossneo (approved):
Reviewed: the animated outlet's hard `height: 100%` was clipping document-style pages so the shell's bottom padding no longer tracked their last content — `minHeight: 100%` with a definite height kept only for Map and Logs is the right minimal fix, and folding `env(safe-area-inset-bottom)` into both the shell clearance and the map's full-bleed negative margin keeps them consistent. Playwright regressions assert the fixed nav no longer intercepts the last touch targets. GUI-only. LGTM.

Nit (non-blocking): docs/screenshots/ is documented as the README/wiki/Pages asset set; PR-evidence shots — especially a 'before' image of the bug — would be better hosted on the PR rather than committed there.

PR #574, opened 2026-09-08
fix(gui): show saved dock position on dashboard map

wjcloudy:
The dashboard's orange dock marker was drawn at the fixed demo position `(0.16, 0.22)`, regardless of the saved dock. Use the dock coordinates from the existing map feed and the same transform as the areas, mower and coverage overlay. Include the dock in the preview bounds and fit the SVG on narrow screens so a dock outside the lawn remains visible.

The GUI backend now omits dock fields until it has received a dock pose, rather than sending an invented `(0, 0)`. An explicitly received origin remains valid. The dashboard hides missing, partial or non-finite coordinates. This only changes the GUI's map view; it does not alter the saved dock or robot navigation.

Validation:

- 12 mocked Playwright checks at 375px and 1440px: saved location, outside-lawn location, valid origin, absent/partial/non-finite pose, alignment with the mower and visible bounds. Screenshots visually reviewed.
- `go test ./...` passed, including dock presence, origin serialization and immutable snapshot regression coverage.
- TypeScript and production frontend build passed.
- Frontend lint: 0 errors (821 existing warnings).
- Full frontend unit suite: 52 files / 389 tests passed.

Based on upstream `dev` at `e252b3d4`, in an isolated bugfix worktree. No mower deployment performed.

### DIFF
--- a/gui/pkg/msgs/mowgli/types.go
+++ b/gui/pkg/msgs/mowgli/types.go
@@ -12,9 +12,10 @@ type Map struct {
 	MapCenterY      float64   `json:"map_center_y"`
 	NavigationAreas []MapArea `json:"navigation_areas"`
 	WorkingArea     []MapArea `json:"working_area"`
-	DockX           float64   `json:"dock_x"`
-	DockY           float64   `json:"dock_y"`
-	DockHeading     float64   `json:"dock_heading"`
+	// Nil means no pose received; an explicitly received zero is a valid dock.
+	DockX       *float64 `json:"dock_x,omitempty"`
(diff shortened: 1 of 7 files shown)


cedbossneo (approved):
Reviewed: Go dock fields become optional pointers so an unset dock is omitted rather than sent as a fake (0,0); the frontend guards missing/partial/non-finite coordinates and MapPage's existing `dock_x !== undefined` check now actually works. Bounds expansion + `preserveAspectRatio=meet` keep an out-of-lawn dock visible. Snapshot test covers the pointer-aliasing trap. GUI-only, no navigation impact. LGTM.

PR #578, opened 2026-09-08
feat(gui): add Diagnostics shortcut to host power controls

wjcloudy:
## Feature

Add a **Host power…** shortcut in **Diagnostics → System**, directly below **Debug Firmware** and above **Last snapshot**. It opens a small menu containing **Restart Host** and **Shut down Host**, making the existing controls easier to find while troubleshooting on desktop or mobile.

The global battery menu remains available. Both entry points use one shared `PowerMenu` component with the existing handlers, confirmation dialogs, backend error notifications, and reboot reconnect/reload flow. The shortcut adds discoverability without introducing a separate power-control implementation.

## Design

- One neutral button opens the menu; selecting an action still requires confirmation.
- The battery menu keeps host and mainboard actions under **Advanced**, alongside its existing **Restart Mowgli** command.
- Reboot retains the blocking reconnect overlay, polling `/api/system/info` and reloading when the GUI returns.
- Shutdown retains the warning that physical access is needed to restore power.
- Both entry points use consistent **Host** wording in English and French.
- Operating-state rules are unchanged: this PR does not add a mowing-state interlock or change the existing backend power policy.

## Installation

No new service, dependency or installer configuration is needed. The existing Linux/systemd execution path uses the GUI container's host PID namespace, privileges and `nsenter`; these are already supplied by the standard installation. The GUI README documents these requirements and how to check older/custom deployments.

## Validation

Existing host reboot validated on Pi 5 hardware. The refactored shortcut was checked with mocked desktop/mobile browser tests, now included in GUI CI; status/translation tests and TypeScript checks pass. No new hardware reboot was performed for this refactor.

### Desktop shortcut

Desktop Host power menu

### Shared reboot confirmation

Reboot confirmation

### Mobile shortcut

Mobile Host power menu

### Shared mobile shutdown confirmation

Mobile shutdown confirmation


cedbossneo (commented):
Review. Not blocking on code quality, but I don't think this should merge as designed:

- **The feature already exists.** `gui/web/src/components/MowerStatus.tsx:152-232` already exposes `POST /system/reboot` and `POST /system/shutdown` ("Restart Pi" / "Shutdown Pi"), mounted globally through `AppShell` so it is reachable from Diagnostics too, with a shared confirmation dialog, a reconnection overlay that polls `/api/system/info` and reloads the page, and the real backend error surfaced (`e.message`). The comment at `MowerStatus.tsx:194-197` records a deliberate decision to tuck destructive actions into an "Avancé" submenu so they are not hit by accident. This PR puts a flat `danger` "Shut down Host" button on the Diagnostics page, with a second implementation of the same flow, a different vocabulary ("Host" / `systemPower.*` vs "Raspberry Pi" / `mowerStatus.*`), no reconnection overlay, and a backend error masked as "request could not be confirmed".
- No disabling while a mow is running (the dialog only warns); no CI job runs the new Playwright spec; branch is 6 commits behind `dev`.

Suggestion: either drop the new component and add a short pointer/link from Diagnostics → System to the existing menu, or refactor so both surfaces share `MowerStatus`'s handlers and dialog. Happy to re-review either way.

wjcloudy:
@cedbossneo I've simplified this in d6816b02 and merged current dev.

Diagnostics now has a single neutral **Host power…** button opening a small menu. The battery menu and this shortcut share the extracted `PowerMenu`: the existing handlers, App-context confirmation dialogs, backend error notifications, and reboot reconnect/reload overlay. The separate timeout/status implementation and flat shutdown button are removed. Both surfaces now say **Host**, with EN/FR translations.

The four mocked browser checks cover the shortcut, battery Advanced menu, confirmation/cancel, reboot polling, and failure recovery. They pass locally and are now wired into GUI CI. Status/translation tests and typecheck pass too; all four PR screenshots are refreshed.

I've kept this scoped to discoverability: the existing operating-state policy is unchanged, so this does **not** add disabling while mowing. A shared backend/UI power interlock would be a separate behaviour change. Could you re-review this simpler version?

cedbossneo (approved):
Re-reviewed d6816b02: one shared PowerMenu (battery menu and Diagnostics shortcut use the same handlers, App-context confirmation dialogs, backend error notifications and the reboot reconnect/reload overlay), neutral button inside a menu, EN/FR at parity, Playwright spec actually executed by GUI CI, dev merged, no conflicts. Approving — thanks for the rework.

Issue #487, opened 2026-08-24
One occupied start pose abandons the whole area ("all N swaths skipped — area not mowable now")


cedbossneo:
Observed on the 2026-08-24 area 1 mow (`dev`). The first start attempt cut **zero grass**; a second attempt 13 minutes later completed the same area at 100 %.

### What happened

The robot undocked and came to rest at ~(4.47, 4.70). Every Nav2 plan from there failed:

```
GridBasedplugin failed to plan from (4.47, 4.70) to (2.07, 9.54): "Start occupied"
```

26 planning calls, all four sub-path starts, then:

```
FollowStrip: segment 1 transit failed — skipping
FollowStrip: segment 2 transit failed — skipping
FollowStrip: segment 3 transit failed — skipping
FollowStrip: segment 4 transit failed — skipping
FollowStrip: all 4 swaths skipped — area 0 not mowable now
```

### The start pose was genuinely inside a keepout

Probed against the live costmap and keepout mask:

| point | keepout_mask | global_costmap |
|---|---|---|
| (4.47, 4.70) — attempt 1 | **100 lethal** | 100 |
| (5.13, 4.21) — attempt 2 | 0 | 0 |

`areas.dat` area_0 has one obstacle: a 33-point circle of **radius 0.25 m centred at (4.06, 4.50)**. The dock sits at ~(6.2, 2.8) and undock reverses on a ~126° bearing — straight at it. Attempt 1 simply stopped further back (see the companion BackUp issue) and landed inside the inflated keepout.

### The gap

SmacPlanner2D has no start tolerance, so a robot standing one cell inside a no-go zone can never plan. `FollowStrip` treats each failed transit as a skipped swath and, once all of them are skipped, declares the area unmowable and ends the run — a whole field forfeited because of a single bad pose.

`coverage_nodes.cpp:806-812` is the skip path.

### Not fixing blind

The obvious recovery — nudge the robot out and retry — means commanding motion on a machine with blades, and the safe direction is not obvious: after undock the robot reversed *into* the spot, so reversing again goes deeper, while mid-mow it arrived driving forward and reversing retraces known-good ground. Clearing costmaps does not help either, since a keepout is a static filter rather than an obstacle layer.

Wanted a maintainer's call on the recovery strategy before writing it.

Found while reviewing a full mow session with Claude Code.
https://claude.ai/code/session_01D46c8dsyRG7k8Q7Djs1NjZ

cedbossneo:
[...]

Issue #446, opened 2026-08-12
Docking calibration fails


fwelvering:
### Component

ROS2 Stack (navigation, SLAM, behavior tree)

### Bug Description

While executing Docking calibration the mower behaves strange and does not complete its calibration.

[YouTube video](https://youtube.com/shorts/wvlc4UF-yvU?is=NrmdGIm-7RUFa85A)

At the end the mower is always about 10cm to far left.

Mower has an good rtk fix with about 34 sattelites. 

### Steps to Reproduce

1. Start Docking calibration

### Hardware & Environment

```markdown
Mower: yardforce classic 500
Gps : um980
Os: Ubuntu
Build: Dev
```

### Relevant Logs

```shell

```

liltaket:
I have experienced the same behavior

fwelvering:
Now the mower gets out of the dock, about 2 meters and than the calibration stops.
Cannot find any relevant logs.

fwelvering:
Unfortunattely, it's still not working.

Endig completely somewhere else and the logs are telling hundreds of times the same warning:
```shell
19:49:51WARN[map_server_node-11] [WARN] [map_server_node]: set_docking_point rejected: EKF yaw not converged (std 1.402Â° > threshold 0.500Â° over 5.0 s, 126 samples). Drive the robot 1 m forward to anchor heading from COG, then retry.
```

fwelvering:
any updates on this?
My mower can't successfully run the calibration and still does not dock after mowing

fwelvering:
unfortunattely, after the latest fix it is still not working.
The mower is about 20cm to far left when it try's to dock.
Still investigating why.

fwelvering:
Follow-up while field-verifying the fix from this issue (#597): the "Start dock calibration" one-click flow could still fail to correct a significantly stale `dock_pose_yaw` in practice, because `docking_server`/`gps_dock_detection_node` only read `dock_pose_x/y/yaw` at container startup — a same-session live re-dock always steered on the OLD heading no matter when the measurement was persisted, so the fix could never verify itself in one run for a heading that was off by more than a small amount.

Opened #621 to persist the measured yaw as soon as it's validated (COG-coherence gate) instead of only after a live re-dock confirms it, and to make the operator-facing message explicit about needing a `mowgli-ros2` restart + one more run to confirm the physical re-dock.

fwelvering:
## Follow-up: root cause found (2 layered bugs), fix scoped

[...]

PR #582, opened 2026-09-10
feat(mowgli_map): keep transit paths a configurable distance from the boundary

fwelvering:
## Summary
- A point-to-point TRANSIT move (planned against the global costmap) had no incentive to stay away from the recorded boundary line — only obstacles and the outside-of-area keepout were costed there. Coverage/mowing itself was never the problem: FTC tracks the F2C path against the LOCAL costmap, which is unaffected by any of this.
- `boundary_inner_margin_m` nudges transit that far inside every recorded edge via a **soft mid-cost penalty** in the global costmap's keepout mask — the same cost tier (`kSoftPenaltyMaskCost`, value 50) the existing outside-slack band already uses. `dock_inner_margin_exempt_radius_m` exempts an isotropic radius around the dock pose from that penalty, so the dock approach carries no bias at all.

## Reworked after review — this is no longer a lethal band

The first version of this PR made the margin **lethal** (mask value 100). Review correctly flagged that as reopening two regressions:

1. **Collides with `chassis_safety_inset`.** Both default to 0.20 m, and the outermost coverage ring is planned exactly `chassis_safety_inset` inside the line — so the ring's own centreline sat inside the newly-lethal-plus-inflated band, reopening the START_OCCUPIED skip cascade (#487) that `chassis_safety_inset` itself was introduced to fix (2026-07-23). **Fixed:** the margin is now a soft cost (~127 after `CostmapFilterInfo`), far below Smac's INSCRIBED(253) cutoff — a start/goal pose in the band never fails "Start occupied", and this is pinned by a new test (`OutermostCoverageRingBandIsNeverLethal`) that asserts the exact cell the outer ring sits on is never lethal.
2. **Walls off narrow area-to-area seams.** A lethal band on both sides of a gap narrower than ~2x the inflated margin blocks it entirely, mowing/navigation areas included. **Fixed:** soft cost never blocks a seam — pinned by a new test (`NarrowSeamBetweenTwoAreasStaysCrossable`, a 0.30 m gap between two areas, narrower than the old 0.40 m inflated lethal band would have allowed).
3. **CI never ran** on the original push (branch/PR created via API tooling rather than the GitHub web flow — first-time-contributor Actions approval was never triggered). Now running via the normal `git push` + `PACKAGES="mowgli_map" ./scripts/test.sh` flow locally; happy to also trigger CI properly if a maintainer can approve the workflow run on the PR.

The rework also makes the ORIGINAL motivation for `dock_inner_margin_exempt_radius_m` (the 2026-04-23 dock-stranding revert, commit 7f4b43d5) moot as a safety concern: a soft-cost cell can never produce "Start occupied" regardless of GNSS drift, dock or not. The exemption is kept anyway — see Docking below.

## Docking, specifically

- **Dock body** (`dock_body_polygon_`, `OBSTACLE_PERMANENT`) stays lethal as before — untouched by this change, the physical charger must stay a real obstacle.
- **Dock corridor** (`dock_corridor_polygon_`, the existing fixed-rectangle carve-out behind the dock body) is unaffected — still forces its cells free regardless of every other pass, exactly as before this PR.
- **`dock_inner_margin_exempt_radius_m`** (new, default 2.5 m) is an isotropic circle around the dock pose, independent of the corridor's fixed orientation/shape, that exempts cells from the soft penalty in *every* direction — including the staging/approach side the robot occupies right after undocking, which the corridor alone does not cover. With the soft-cost design this is no longer load-bearing for safety (see above), but it keeps the dock approach completely bias-free rather than merely "never blocked", which still matters given `.claude/rules/ros2.md`'s note that the dock canopy commonly pins GNSS at Float.
- Sim no longer force-disables the margin (dropped the `boundary_inner_margin_m: 0.0` override in `sim_full_system.launch.py`): that override existed only to dodge the old lethal design's "goal occupied" risk in the compact test polygon. The soft-cost design carries no such risk — it now runs in sim exactly like the pre-existing `enforce_boundary_margin_m` soft band already does, unconditionally.

## Test plan
- [ ] `cd ros2 && PACKAGES="mowgli_map" ./scripts/test.sh` — `BoundaryInnerMarginTest` grows from 5 to 7 cases: interior stays fully free; an edge far from the dock gets the soft penalty, never lethal; the outermost coverage ring's own band is never lethal (the exact chassis_safety_inset collision above); the same edge distance near the dock carries zero penalty; a 0.30 m seam between two adjacent areas stays crossable; disabling the dock exemption falls back to the soft penalty (still never lethal); disabling the margin entirely restores the pre-existing behaviour.
- [ ] `python3 ros2/scripts/check_config_drift.py` (already run locally, clean)
- [ ] `cd gui/web && npx tsc --noEmit && yarn lint && yarn test` (schema/template defaults kept in lockstep for the generic Go parity test; i18n en/fr key parity verified locally)
- [ ] **Field test blocked** by #446 (dock calibration currently fails / mower won't dock reliably on this operator's hardware) — can't yet verify live that a normal undock/dock cycle succeeds with this change applied. The soft-cost design and the dock exemption geometry are both covered in isolation by `BoundaryInnerMarginTest` (no live docking needed), but end-to-end field/sim validation with a real dock pose remains outstanding until #446 is resolved. Reviewers/maintainers with a working dock are encouraged to field-test before merging.


cedbossneo (changes requested):
Blocking, sorry — this reopens a regression we already paid for on 2026-07-23.

1. `boundary_inner_margin_m` was deliberately wired to `0.0`; this PR turns it back on at **0.20 m**, which equals `chassis_safety_inset`. The outer coverage ring (headland ring 0, and the outermost swath when rings are off) is planned exactly `chassis_safety_inset` inside the line, so it now sits **in the new lethal band**; after inflation the INSCRIBED band reaches ~0.40 m inside the boundary. Consequences we have measured before: SmacPlanner2D has no start tolerance, so every blade-off transit from the edge answers `Start occupied` (the START_OCCUPIED skip cascade, CLAUDE.md Invariant 7 / #487), and `detour_resume` re-confirms "obstacle" on every abort. The PR text says it makes edge transits *stop* failing "Start occupied"; the geometry says the opposite for our stock config.
2. The same band is painted on every seam between adjacent areas, navigation areas included: a 0.40 m lethal wall between two areas the robot must cross.
3. CI has never run on this branch.

If the goal is to keep *transit* paths (RPP/Smac) away from the edge without touching *coverage*, the lever is a transit-only cost (global costmap inflation / a mid-cost band like `kOutsideSlackMaskCost`), never a lethal band inside the polygon, and it must leave the dock approach and area seams free. Happy to re-review a version along those lines.

fwelvering:
The thing i tryed to fix with this is the mower constantly transitting just outside the boundry and gets stuck.
Manual intervention is than needed to get it running again.

Will give this some thoughts.

cedbossneo (approved):
Re-reviewed 25410373: the band is no longer lethal (shared mid cost ~127 via the KeepoutFilter mapping), the inner band is exempted around the dock (dock_inner_margin_exempt_radius_m 2.5), and the three blockers each have a named regression test. Approving.

One follow-up worth a look (not blocking): inner and outer bands now share the same constant (`kSoftPenaltyMaskCost`, `costmap_filters.cpp:271/275`). Before, 0.1 m inside cost 0 and 0.1 m outside cost 127, so a constrained planner still had a reason to stay inside; with equal costs, cutting a corner through the outside band is free in relative terms — the opposite of the symptom you were fixing. A slightly lower inner cost (e.g. 80 vs 127) would keep the gradient pointing inward.

PR #584, opened 2026-09-11
fix(behavior): keep HL state AUTONOMOUS while docking after a completed mow

wjcloudy:
## Summary

A completed mow leaves the robot frozen where it stands: it reports `MOWING_COMPLETE`, never drives to the dock, and accepts no further command until the stack is restarted (which does not help) or the operator drives it in manual mode.

The tree publishes `HighLevelStatus state=1` (`HIGH_LEVEL_STATE_IDLE`) and *then* runs `DockRobot`. `hardware_bridge` forwards that state verbatim to the STM32 (`current_mode_ = msg->state`), where `HL_MODE_IDLE` becomes `OPENMOWER_STATUS_IDLE` and `motors_handler()` re-asserts it as a per-cycle hard stop:

```c
} else if (main_eOpenmowerStatus == OPENMOWER_STATUS_IDLE) {
  hard_stop = true;
```

So the tree declares the robot idle and then asks a hard-stopped robot to drive to the dock.

## Evidence from the robot

Captured live, robot 6.4 m from the dock:

- `/cmd_vel_nav` publishing at 10 Hz, `/cmd_vel` (post-mux) carrying the command
- encoder totals frozen at `L=26704 R=23167` for **474 consecutive samples**, not one tick
- no emergency, no dig latch, battery 27.0 V, firmware healthy and reporting
- Nav2 looping `follow_path` → `Failed to make progress` → `spin` → `backup`, all timing out
- driving `/cmd_vel_tuning` (priority 30, bypassing Nav2 entirely) at 0.15 m/s also produced **zero** motion, confirming the block is at the wire, not in planning

Manual mode recovered it, because `HL_MODE_MANUAL_MOWING` maps to `OPENMOWER_STATUS_MOWING`.

## Why now

The IDLE publish predates the gate and was latent and harmless. `a8fbe5fd` ("discontinuous coverage + per-swath MPPI, fusion/docking fixes") added two IDLE gates — one in `on_cmd_blade` for blade safety, and one in `motors_handler()` that also hard-stops the **wheels**. That turned a dormant inconsistency into a deadlock, with no tree change to blame.

## The fix

Publish `state=2` (AUTONOMOUS) with `state_name` unchanged at `MOWING_COMPLETE` until the robot is actually docked. The existing `IDLE_DOCKED` publish after `DockRobot` remains the real idle transition.

This is fixed in the tree rather than in firmware on purpose: firmware must stay free to refuse motion while it believes it is idle or docked.

**All 7 `DockRobot` sites were audited.** Five already publish `state=2` before docking — `CRITICAL_BATTERY_DOCKING`, `RAIN_DETECTED_DOCKING`, `LOW_BATTERY_DOCKING`, `COVERAGE_FAILED_DOCKING`, `RETURNING_HOME`. Only the two `MOWING_COMPLETE` sites were outliers, so this makes them consistent with five existing precedents rather than introducing a new convention.

## Regression test

`test_dock_motion_gate.cpp` walks the real `main_tree.xml` and asserts the last state published before any `DockRobot` is not IDLE. Structural rather than behavioural on purpose: the defect is a mismatch between a literal in the tree and a gate in firmware, so ticking mocked nodes cannot see it — the tree "succeeds" while the wheels never turn.

Verified to discriminate: fails on the pre-fix tree naming both offending sites, passes on the fixed tree.

## Consumer review

- `scheduler.go safeToStart()` — **behaviour preserved** (third commit). Raising the state to AUTONOMOUS would otherwise make the scheduler skip a mow falling due while the robot trundles home after a finished run, which used to fire. `state_name` is now tracked alongside the numeric state and `MOWING_COMPLETE` is exempted — and only that one. `RETURNING_HOME`, `LOW_BATTERY_DOCKING`, `CRITICAL_BATTERY_DOCKING` and `RAIN_DETECTED_DOCKING` stay blocked, as they were before this series: a schedule must not interrupt the robot protecting itself. `safeToStart()` has a single caller and gates only *scheduled* starts — the GUI Start button is unaffected either way
- `session_tracker.go` — unaffected, keys on `state_name`
- idle debounce in `status_nodes.hpp` — unaffected; debounce state is per-node-instance, so `IDLE_DOCKED` keeps its own history and still publishes immediately
- `led_ring_node.cpp` — cosmetic only: shows autonomous during the dock transit, exactly as `RETURNING_HOME` already does
- blade safety — `SetMowerEnabled(false)` runs *before* the publish in both sequences, so lifting the IDLE blade gate cannot arm blades

## Test status

**Confirmed end to end on the robot.** Full multi-area run to completion, on the patched tree:

```
mode=2 (AUTONOMOUS), state_name='MOWING_COMPLETE'   <- was mode=1 (IDLE)
docking_server: Attempting to dock robot at home_dock.
docking_server: Successful navigation to staging pose      (+56 s of driving)
docking_server: Successful initial dock detection
docking_server: Made contact with dock
docking_server: Robot is charging!
mode=1 (IDLE), state_name='IDLE_DOCKED'
EndSession: clearing per-session flags
```

79 s from mow complete to charging, against an indefinite deadlock before.
Encoders ran continuously through the transit (`cum L=342057 -> 342338` in
3 s) where they had previously sat frozen at `L=26704 R=23167` for 474
consecutive samples.

Note the last two lines: `IDLE_DOCKED` correctly drops back to `mode=1`
once the robot is actually docked, so the firmware idle gate is restored
exactly where it belongs — the change moves that transition, it does not
remove it.

### DIFF
--- a/gui/pkg/providers/scheduler.go
+++ b/gui/pkg/providers/scheduler.go
@@ -39,9 +39,10 @@ type SchedulerProvider struct {
 	dbProvider   types.IDBProvider
 	soilProvider types.ISoilProvider
 
-	mu                 sync.RWMutex
-	lastHighLevelState uint8
-	lastEmergency      bool
+	mu                     sync.RWMutex
+	lastHighLevelState     uint8
+	lastHighLevelStateName string
+	lastEmergency          bool
 }
 
 // NewSchedulerProvider creates and starts the scheduler background goroutine.
@@ -76,9 +77,17 @@ func (s *SchedulerProvider) subscribeToStatus() {
 					break
 				}
 			}
+			// Accept "state_name" or "StateName"
+			for _, k := range []string{"state_name", "StateName"} {
+				if v, ok := raw[k]; ok {
+					_ = json.Unmarshal(v, &hls.StateName)
+					break
+				}
+			}
 		}
 		s.mu.Lock()
 		s.lastHighLevelState = hls.State
+		s.lastHighLevelStateName = hls.StateName
 		s.mu.Unlock()
 	}); err != nil {
 		logrus.Warnf("Scheduler: failed to subscribe to highLevelStatus: %v", err)
@@ -215,7 +224,9 @@ func (s *SchedulerProvider) persistSkip(sched schedule, reason string, now time.
 // safeToStart returns true when it is safe to send COMMAND_START.
 // It blocks mowing when:
 //   - an emergency is active (latched or active), or
-//   - the robot is already in autonomous (2) or recording (3) state.
+//   - the robot is already in autonomous (2) or recording (3) state,
+//     EXCEPT the post-mow dock transit (state 2, state_name MOWING_COMPLETE),
+//     which reports autonomous only to keep the firmware wheel gate open.
 //
 // HIGH_LEVEL_STATE constants:
 //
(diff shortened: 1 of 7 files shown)


cedbossneo (commented):
Verified the diagnosis end to end (HighLevelStatus.state → hardware_bridge `current_mode_` → firmware `motors_handler` hard-stop on IDLE) and the tree change against the other docking sites; the structural test does check the tree that is actually loaded, and CI is green.

One correction was needed, pushed to this branch as `53fda816` (maintainer edit): the description says `session_tracker.go` "keys on state_name" — it keys on the numeric state (`gui/pkg/providers/session_tracker.go:195`, `isMowing := status.State == 2`). With state staying at 2 during the return trip, the mowing session no longer closed at MOWING_COMPLETE: distance/duration absorbed the return-to-dock trip, and a docking failure after a finished mow rewrote the completed session as `error`. Excluding MOWING_COMPLETE from `isMowing` restores the previous session boundaries; regression test added (`TestSessionTracker_MowingCompleteWithAutonomousStateEndsTheSession`).

Non-blocking follow-ups: the regex scan in `test_dock_motion_gate.cpp:94-106` is line-based, so a multi-line `<PublishHighLevelStatus …>` before a `DockRobot` (the file already has three multi-line publishes) would silently escape it — a `sregex_iterator` over the whole text plus an "every tag matched" assertion would make that loud; `docs/claude/codemaps/mowgli_behavior.md` should mention the AUTONOMOUS-while-docking rule.

Will approve once CI is green on `53fda816`.

cedbossneo (approved):
CI green on 53fda816 (session_tracker fix included). Diagnosis, tree change and test verified; merging.

PR #585, opened 2026-09-11
fix(foxglove): bundle Universal GNSS interface typesupport

### LINKED ISSUE #581: foxglove_bridge: "package 'universal_gnss_ros2' not found" — package missing from mowgli-ros2:dev image
### Component

Docker / Deployment

### Bug Description

Expected: foxglove_bridge should either successfully advertise the
`/universal_gnss_receiver/get_snapshot` service, or not attempt to at all
if it's meant to stay internal (it's already excluded from
topic_whitelist/client_topic_whitelist in foxglove_bridge.yaml).

Actual: foxglove_bridge in the mowgli-ros2 container repeatedly logs:

  [foxglove_bridge]: Failed to add service "/universal_gnss_receiver/get_snapshot"
  (universal_gnss_ros2/srv/GetReceiverSnapshot): package 'universal_gnss_ros2' not found

...once per reconnect, spamming the logs during normal mowing operation.

### Steps to Reproduce

1. Run the standard compose stack (mowgli-ros2 + mowgli-gps sidecar),
   both on the `:dev` image tag, with GNSS_BACKEND=universal,
   GPS_PROTOCOL=UBX, GNSS_STACK=universal (mowgli-gps.yml default).
2. Watch `docker logs mowgli-ros2` — the WARN above appears repeatedly
   (roughly every reconnect / foxglove client (re)connection).
3. Confirm inside the container:

   docker exec mowgli-ros2 bash -lc '
     source /opt/ros/kilted/setup.bash
     [ -f /opt/ublox_msgs/setup.bash ] && source /opt/ublox_msgs/setup.bash
     [ -f /ros2_ws/install/setup.bash ] && source /ros2_ws/install/setup.bash
     ros2 pkg list | grep -i gnss
     ros2 pkg prefix universal_gnss_ros2
   '

   -> "gnss packages" grep returns nothing, `ros2 pkg prefix` returns
   "Package not found".

4. Compare with the GPS sidecar, where the package IS present and its
   receiver_node (rclcpp node name "universal_gnss_receiver") is the one
   actually advertising the service on the shared ROS graph:

   docker exec mowgli-gps bash -lc '
     source /opt/ros/kilted/setup.bash
     ros2 pkg list | grep -i gnss
   '

### Hardware & Environment

```markdown
- Images: ghcr.io/mowglinext/mowglinext/mowgli-ros2:dev,
          ghcr.io/mowglinext/mowglinext/gps:dev
- Env: GNSS_BACKEND=universal, GPS_PROTOCOL=UBX, GNSS_STACK=universal
- Containers observed: mowgli-ros2, mowgli-gps, mowgli-lidar, mowgli-gui,
  mowgli-mqtt, mowgli-watchtower (all "Up", no restarts/crashes)
```

### Relevant Logs

```shell
[foxglove_bridge-18] [WARN] [1789028913.286547928] [foxglove_bridge]: Failed to add service "/universal_gnss_receiver/get_snapshot" (universal_gnss_ros2/srv/GetReceiverSnapshot): package 'universal_gnss_ros2' not found
[foxglove_bridge-18] [WARN] [1789028913.793989097] [foxglove_bridge]: Failed to add service "/universal_gnss_receiver/get_snapshot" (universal_gnss_ros2/srv/GetReceiverSnapshot): package 'universal_gnss_ros2' not found
(repeats on every reconnect)
```

ros2/Dockerfile's `deps` and `build` stages appear to include
`ros2/src/external/universal-gnss/gnss_ros2` (universal_gnss_ros2) in the
plain workspace-wide `colcon build` — no COLCON_IGNORE or
--packages-skip was found for it — so the package's typesupport should in
theory end up in /ros2_ws/install and be sourced by ros2_entrypoint.sh.
Yet it's verifiably absent from the running `:dev` image. Worth checking
the ros2-docker.yml build log for that image tag to see whether the
`universal-gnss` submodule was actually checked out for that specific
build, or whether `colcon build` silently skipped/failed the package
(the `deps` stage's `rosdep install ... || true` would swallow an
unresolved dependency for it).

This is the same failure mode already hit (and fixed) for a different
vendored driver: SESSION-2026-04-18.md documents the identical symptom
for /ubx_* topics ("schema not found... the container doesn't have
ublox_ubx_msgs installed — only the GPS container does"), which was
later fixed by adding a dedicated `ublox-msgs-builder` stage to
ros2/Dockerfile that builds *only* the interface packages into the
mowgli-ros2 image so foxglove_bridge can resolve schemas without needing
the full driver. No equivalent stage/fix exists for universal_gnss_ros2.

Two possible fixes:
1. Root-cause fix: figure out why universal_gnss_ros2 isn't ending up in
   /ros2_ws/install in the mowgli-ros2 image despite appearing to be
   part of the full workspace build, and correct that.
2. Same pattern as ublox: add a slim stage that builds only
   universal_gnss_ros2's msg/srv interfaces into mowgli-ros2.
3. Quick mitigation (doesn't fix the underlying gap, just silences the
   log): add a `service_whitelist` to
   ros2/src/mowgli_bringup/config/foxglove_bridge.yaml excluding
   `/universal_gnss_receiver/.*`, mirroring the existing
   topic_whitelist/client_topic_whitelist regex that already hides
   `/universal_gnss/*` topics.

Not mowing-safety-critical — /gps/fix, /gps/status and /rtcm are
published fine by mowgli_gnss_bridge in the GPS sidecar regardless of
this. It's purely a foxglove_bridge introspection/log-noise issue.

liltaket:
## Summary

Provide `universal_gnss_ros2` message and service typesupport in the final ROS2 runtime image so foxglove_bridge can introspect the GNSS sidecar service without shipping a second GNSS driver runtime.

Fixes #581.

## Root cause

The recursive checkout is complete and the pinned submodule is present, but colcon identifies `ros2/src/external/universal-gnss` as the top-level generic CMake project `universal_gnss`. Package discovery then stops before the nested `gnss_ros2/package.xml`. The generic project builds successfully with `UNIVERSAL_GNSS_BUILD_ROS2=OFF`, so `/ros2_ws/install` contains Universal GNSS tools but no `universal_gnss_ros2` package or generated interfaces.

This was confirmed in the current `dev` amd64 build log and published image layer: the build starts and installs `universal_gnss`, while the final install tree has no `universal_gnss_ros2` ament index entry. The normal entrypoint does source `/ros2_ws/install`, so the defect occurs before runtime sourcing.

## Changes

- Build only the pinned Universal GNSS `.msg` and `.srv` definitions in a dedicated overlay.
- Fail the overlay build if the pinned submodule interface set differs from the single explicit `rosidl_generate_interfaces` list, with an actionable update message.
- Ignore the submodule root during the main workspace build, avoiding the unintended driver and tool install.
- Copy and explicitly source the interface overlay before the main workspace overlay in the final runtime image.
- Verify `ros2 pkg prefix universal_gnss_ros2` plus `GnssStatus`, `RtcmFrame`, and `GetReceiverSnapshot` through `ros2 interface show` against each pushed final digest on amd64 and arm64.

The interface builder copies its manifest and definitions directly from the same superproject-pinned submodule used by the GPS image, so Docker cache invalidation and schema revision stay tied to the gitlink. The drift guard derives its discovered set from those copied sources and compares it with the same list passed to `rosidl_generate_interfaces`, avoiding a second hidden expected list.

## Foxglove test boundary

A protocol-level Foxglove WebSocket test was assessed but not added. It would require starting the bridge and a temporary service plus maintaining a Foxglove protocol client not otherwise present in the runtime image. That is disproportionate and brittle for this focused fix. The final-image interface smoke test is the selected deterministic boundary: it verifies package discovery and generated ROS introspection/typesupport for every bundled Universal GNSS interface without starting unrelated robot services.

## Validation

- Positive isolated CMake configure of the exact three pinned interfaces
- Negative drift-guard configure with an added interface: failed with the drifted filename, both interface sets, and the required CMake file update
- `bash -n ros2/scripts/ros2_entrypoint.sh`
- Parsed `.github/workflows/ros2-docker.yml` with Ruby YAML
- Verified the pinned package name and exact copied interface inputs
- Verified no Mowgli package declares a build dependency on the ignored Universal GNSS runtime package
- Reviewed overlay sourcing order and the interface-only Docker stage boundary
- `git diff --check`
- Required [ROS2 Build and Test](https://github.com/mowglinext/mowglinext/actions/runs/34656511664): passed
- [Cached native image matrix](https://github.com/liltaket/mowglinext-lab/actions/runs/34656505570): amd64 and arm64 builds, all final-digest package/interface smoke tests, and manifest merge passed
- [Clean native image matrix](https://github.com/liltaket/mowglinext-lab/actions/runs/34656776529): amd64 and arm64 builds, all final-digest package/interface smoke tests, and manifest merge passed
- GPS sidecar image workflow: passed with the unchanged pinned submodule revision

The local Docker daemon was stopped, so it was not restarted. Native GitHub runners performed both clean and cached final-image validation. Hardware testing is not required because this changes schema availability only; motion, blade control, GNSS ownership, and runtime startup commands are unchanged.

## Component

- [x] ROS2 Stack (`ros2/`)
- [x] CI/CD (`.github/`)

## Checklist

- [x] No unrelated changes
- [x] Physical behavior is unaffected
- [x] No C++ source or interface definitions changed


cedbossneo (commented):
Reviewed against `dev`: universal-gnss submodule gitlink is unchanged (no silent re-pin), zero conflicts, the anti-drift guard fails closed. Only caveat: `ros2-docker` has no `pull_request` trigger, so the bundled image is validated only after merge — watch the first `dev` build. Looks good to merge.

cedbossneo (approved):
Submodule gitlink unchanged, zero conflicts, fail-closed guard; CI green. Merging.

Issue #583, opened 2026-09-10
Z-axis accelerometer reads ~33% high at rest (13.09 vs 9.81 m/s²) — permanently stalls costmap_scan_filter's ground filter


fwelvering:
### Component

Hardware Bridge / Firmware

### Bug Description

Expected: At rest, /imu/data's linear_acceleration should read ~9.81 m/s²
total magnitude (gravity), close enough to the hardcoded kGravityMs2
(9.81) in costmap_scan_filter_node's ground-filter outlier guard that
fresh samples get accepted and the tilt-correction (slope ground-return
filtering) actually runs.

Actual: On this robot, linear_acceleration.z reads a very stable ~13.09
m/s² at rest (measured over 15 samples with the blade OFF, robot
stationary: min 13.068, max 13.113, mean 13.089 — a spread of only
~0.35%, i.e. not vibration noise). x/y are correctly near 0. That's a
consistent +3.28 m/s² / ~1.335x overshoot vs. the expected 9.81.

Consequence: costmap_scan_filter_node's ground filter (the code that
corrects LiDAR ground-slope returns using the IMU gravity vector, see
its own file-header comment on why this exists — without it, real
ground returns at 1-2m show up as walls and the planner refuses to
drive) computes |accel| = sqrt(ax²+ay²+az²) ≈ |az| ≈ 13.09 when level,
compares it to 9.81 ± accel_g_tolerance_ms2 (default 3.0), and — since
13.09 - 9.81 = 3.28 > 3.0 — treats essentially every sample as a
"motion-dominated outlier" and rejects it. Because rejection freezes
BOTH the cached gravity vector AND the "last fresh sample" timestamp
(it only compares future samples against that frozen value, never
against a fixed reference), this node can get permanently stuck once
it first trips: on this robot it logged "ground filter idle: last IMU
sample Ns old (>0.50s)" 8397 times in 24h, with the tail of that run
showing an UNBROKEN, still-growing gap (4950s → 4970s, in lockstep
5.0s throttle steps) — i.e. the ground filter had not accepted a
single fresh sample in at least ~83 minutes straight at capture time.
Net effect: the slope/ground-tilt correction this node exists for has
likely been non-functional for the large majority of this robot's
recent mowing time, independent of any actual terrain slope.

### Steps to Reproduce

1. With the robot powered on and stationary, blade OFF:

   docker exec mowgli-ros2 bash -lc '
     source /opt/ros/kilted/setup.bash
[...]

PR #586, opened 2026-09-12
fix: recover ground filter from stable IMU offset

liltaket:
## Summary

- extract the costmap ground filter's acceleration gate into a directly tested pure helper
- preserve the standard-gravity fast path and stale/pass-through safety behavior
- allow a plausible out-of-band baseline to re-seed only after five seconds of continuous, full-vector consistency
- restart the recovery window after stale sample gaps, clock rollback, invalid input, candidate inconsistency, or a normal accepted sample
- require enough observations for the configured window and reject unsafe timing configurations
- limit replacement gravity directions to 30° tilt so a frozen off-axis bus remains fail-safe
- warn on an actual re-seed with the old/new baseline and absolute/percentage change
- permanently reject non-finite values and magnitudes outside 0.5g..1.5g

Addresses the ground-filter stale-latch portion of #583. It does not close the underlying WT901 scale/range investigation.

## What the investigation found

Before this PR, the node accepted its first nonzero acceleration sample even when its magnitude was outside the configured gravity tolerance. Once it had an estimate, later out-of-band samples were rejected without refreshing the freshness timestamp. That correctly made the ground filter fall back to conservative pass-through after `imu_max_age_s`, but a stable 13.09 m/s² stream could never make the ground filter usable again.

The persisted hardware-bridge calibration does not currently contain `mean_az` or a gravity-magnitude baseline. Its Z mean is only a local value used to report implied mounting tilt, so this change does not reinterpret that file or couple the costmap node to it.

The WT901 driver still assumes an accelerometer range without configuring or reading it back. The observed approximately 1.335x ratio is not an obvious standard full-scale mismatch, and there is not enough repository or hardware evidence to change that conversion in this PR.

Recommended follow-up scope is WT901 configured-range verification, scale-factor investigation, and device-to-device comparison. No follow-up issue is created by this PR.

## Safety behavior

Rejected and invalid samples do not refresh IMU freshness. Until a new baseline proves stable, `/scan_costmap` falls back to pass-through. `/scan_collision` remains outside the gravity filter and is unchanged.

A recovery candidate must remain within 0.5 m/s² of its initial full acceleration vector for five continuous seconds. The maximum gap between candidate samples is the existing `imu_max_age_s` value (default 0.5 s). Larger gaps and backwards time restart the full window. Re-seed requires both the full elapsed duration and at least `ceil(reseed_after_s / candidate_max_gap_s) + 1` consistent observations. Non-finite/non-positive timing and a maximum gap large enough to permit a two-sample window disable recovery.

Before replacing the trusted direction, the estimator also requires the candidate mean to be within `max_reseed_tilt_rad` (default 30°) of vertical. An off-axis candidate above that bound is cleared and remains `REJECTED`; it cannot change the direction/baseline or refresh IMU freshness. `/scan_costmap` therefore continues toward stale-data pass-through rather than removing obstacles from an untrusted gravity estimate.

Using a fixed anchor prevents a slowly drifting sequence from following its own moving mean. Values outside 0.5g..1.5g never seed or re-seed the filter.

Strong blade vibration may delay recovery until the vector settles; this is intentionally conservative and this PR does not attempt a general vibration-filter redesign.

## Tests

Focused pure-state regression coverage includes:

- standard gravity, immediate normal seeding, and low-pass tilt acceptance
- startup and recovery on a continuously stable 13.09 m/s² offset
- stale gaps not counting toward the five-second window
- a partway gap requiring a fresh full window
- clock rollback restarting the candidate
- exact maximum-gap and equal-timestamp behavior
- elapsed time alone and permissive/invalid timing configurations cannot re-seed
- safe approximately 23° tilted recovery and the exact 30° boundary
- just-over-limit and approximately 57° frozen off-axis candidates remain rejected across repeated windows
- fixed-anchor inconsistent and slowly drifting candidates
- invalid/non-finite input and accepted samples clearing candidates
- exact 0.5g and 1.5g plausibility boundaries

Local checks:

- Xcode clang-format dry run: pass
- standalone C++17 compile/runtime continuity scenarios: pass
- `git diff --check`: pass

The macOS host has no local ROS Kilted installation, and Docker Desktop is not running, so the ROS package build and ament/gtest suite are validated by GitHub CI.

Fresh GitHub CI at final rebased head `6d4d5d4e`:

- ROS Kilted full workspace build: pass
- `mowgli_localization`: 14/14 tests passed, including `test_gravity_estimator`
- full ROS test summary: 1810 tests, 0 errors, 0 failures
- clang-format, cppcheck, config drift, and external GNSS launch contract: pass

## Validation boundary

No mower or WT901 hardware was available. This PR fixes the demonstrated costmap ground-filter stale-latch/recovery behavior only. It does not establish why the WT901 reports approximately 13.09 m/s² instead of approximately 9.81 m/s², change the firmware conversion, or claim hardware validation.

### DIFF
--- a/ros2/src/mowgli_localization/CMakeLists.txt
+++ b/ros2/src/mowgli_localization/CMakeLists.txt
@@ -184,6 +184,8 @@ add_executable(costmap_scan_filter_node
   src/costmap_scan_filter_node.cpp
 )
 
+target_include_directories(costmap_scan_filter_node PUBLIC ${INCLUDE_DIR})
+
 ament_target_dependencies(costmap_scan_filter_node
   rclcpp
   sensor_msgs
--- a/ros2/src/mowgli_localization/CMakeLists.txt
+++ b/ros2/src/mowgli_localization/CMakeLists.txt
@@ -380,6 +380,15 @@ if(BUILD_TESTING)
     ${CMAKE_CURRENT_SOURCE_DIR}/include
   )
 
+  # ---- test_gravity_estimator ----
+  # Pure C++ regression test for the costmap IMU gravity gate.
+  ament_add_gtest(test_gravity_estimator
+    test/test_gravity_estimator.cpp
+  )
+  target_include_directories(test_gravity_estimator PRIVATE
+    ${CMAKE_CURRENT_SOURCE_DIR}/include
+  )
+
   # ---- test_robot_yaml_scalar ----
   # Guards the ONE canonical dock_pose persist path (Invariant 6): the
   # motion-derived (yaw_source=MOTION) write must splice only the dock_pose_*
--- a/ros2/src/mowgli_localization/include/mowgli_localization/gravity_estimator.hpp
+++ b/ros2/src/mowgli_localization/include/mowgli_localization/gravity_estimator.hpp
@@ -18,11 +18,13 @@
 
 #include <cmath>
 #include <cstddef>
+#include <limits>
 
 namespace mowgli_localization
 {
 
 constexpr double kStandardGravityMs2 = 9.80665;
+constexpr double kDefaultMaxReseedTiltRad = 0.52359877559829887308;
 
 struct GravityVector
 {
@@ -44,6 +46,8 @@ struct GravityEstimatorConfig
   double candidate_max_gap_s{0.5};
   /// Stable candidate duration needed to replace a stale baseline.
   double reseed_after_s{5.0};
+  /// Largest tilt allowed when adopting a replacement gravity direction.
+  double max_reseed_tilt_rad{kDefaultMaxReseedTiltRad};
   /// Direction low-pass weight for ordinary accepted samples.
   double direction_alpha{0.2};
 };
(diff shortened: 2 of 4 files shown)


cedbossneo (commented):
Draft feedback — the direction is right (no impact on lift/tilt thresholds, tests present), one thing must change before this is safe:
- `gravity_estimator.hpp:104-111`: the re-seed accepts **any** stable mean as the new gravity direction, with no tilt bound. A bus frozen at an off-axis value for 5 s re-seeds the ground filter with an implicit tilt of tens of degrees, turning a fail-safe into a fail-active that can erase real obstacles from `/scan_costmap`. Add `max_reseed_tilt_rad` (~0.52 rad / 30°, above the steepest mowable slope) and refuse the re-seed when `hypot(mean.x, mean.y)/|mean| > sin(max_reseed_tilt)`; ~10 lines + 2 tests.
- The "5 s continuous" window is not actually guaranteed by the sample counter as written.
Then it is mergeable.

liltaket:
Thanks for catching this safety issue. Fixed in 6d4d5d4e: max_reseed_tilt_rad now defaults to 30 degrees, and an over-limit candidate is cleared and remains REJECTED, so it cannot replace the baseline/direction or refresh IMU freshness. Reseed now requires valid timing configuration, the full elapsed window, every gap within candidate_max_gap_s, and at least ceil(reseed_after_s / candidate_max_gap_s) + 1 consistent samples; configurations that could permit a two-sample multi-second window disable recovery. Regression tests cover safe and exact-boundary tilt, just-over-limit and frozen off-axis rejection across repeated windows, sparse/invalid timing, and the existing gap/rollback/equal-time cases. The branch is rebased onto current dev (52b902b0), and fresh CI is green: mowgli_localization 14/14 and 1810 tests overall with 0 errors/failures.

cedbossneo (approved):
Re-reviewed 6d4d5d4e: max_reseed_tilt_rad (30° default, bounded ≤ π/2), over-limit candidates rejected without refreshing IMU freshness, continuous-window requirement with sample-count floor, regression tests for boundary/over-limit/frozen-off-axis/sparse timing; no effect on tilt/lift thresholds, no new ROS parameter, CI green. Approving.

PR #579, opened 2026-09-09
feat(mowgli_leds): dim status ring after charge is complete and idle

fwelvering:
Closes #550.

## Summary
- A fully charged robot left on the dock currently shows the WS2812 status ring at steady full-green brightness indefinitely, which is bright enough at night to bother nearby people (issue #550).
- Add `led_charge_complete_timeout_s` (default `600.0`s / 10 min, `0` disables the feature) and `led_charge_complete_dim_scale` (default `0.0`, i.e. off) so the ring dims itself after that long at "charge complete", and snaps back to full brightness immediately if the robot is unplugged or the level drops again.
- Wired through the pure pattern logic (`led_pattern.hpp`), the ROS2 node (`led_ring_node.hpp/.cpp`), the bringup template + launch passthrough, and the GUI (schema, Status LEDs settings section, en/fr i18n), following the existing `led_*` parameter conventions documented in the package's codemap.

## Test plan
- [x] `colcon build --packages-select mowgli_leds && colcon test --packages-select mowgli_leds && colcon test-result --verbose` (4 new gtest cases added to `test_led_pattern.cpp`, 32 -> 36)
- [x] `cd gui && go test ./pkg/api/...` (`settings_leds_test.go` expected-key list updated)
- [x] `cd gui/web && npx tsc --noEmit && yarn lint && yarn test` (`LedsSection.test.tsx` updated)
- [x] `python3 ros2/scripts/check_config_drift.py` (already run locally, clean)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #550, opened 2026-09-06
Ledring disable on idle


fwelvering:
### Component

ROS2 Stack

### Problem or Motivation

Now when the robot is on it's dock and the battery is fulle charged the green ring gives a lot of light.

### Proposed Solution

When it's on the dock and charging is finished disable the leds after x (configurable) minutes.
Maybe let a single configurable led ID (or few id's) on to see it's state.

### Alternatives Considered

Not using the led ring, but i think it is a good addition to the mower.

### Would you work on this?

None

PR #596, opened 2026-09-13
fix(mowgli_leds): keep select indicator pixels lit after charge-complete dim

fwelvering:
## Summary

This is a bug fix, not a new feature: issue #550 and the description of #579 (merged) both promised *"optionally keeping select LED indicators active to display operational status"* once the ring dims after sitting fully charged for a while. #579 only shipped a uniform whole-ring dim (`led_charge_complete_dim_scale`) — the "select indicators" half of the original ask was never actually built, and wasn't exposed in the GUI either.

## What's added

Two ways to keep a handful of pixels lit at a distinct brightness while the rest of the ring dims to `led_charge_complete_dim_scale`:
- **`led_charge_complete_indicator_count`** — N pixels, evenly spaced around the ring (e.g. 4 = every 90°). `0` (default) disables this — behaviour is unchanged from #579 unless an operator opts in.
- **`led_charge_complete_indicator_ids`** — exact pixel indices instead, e.g. `"0,4,8,12"`, for an operator who wants a specific pixel ("the one facing the house") rather than an auto-computed spacing. Takes priority over the count when non-empty.

`led_charge_complete_indicator_ids` is a comma-separated **string**, not a native ROS2 array: an empty array default can't be type-inferred in a params YAML and fails to load at startup — the same gotcha `mowgli_map`'s `area_names` already works around (see `mowgli_map/config/map_server.yaml`). Parsed by a small `ParseIndicatorIds()` helper in `led_ring_node.cpp` that silently skips malformed/negative tokens rather than rejecting the whole list.

Both knobs are independent of `led_charge_complete_dim_scale` (now governs the non-indicator background pixels only) and are exposed in **Settings → Status LEDs** alongside the existing charge-complete fields, following the same template → launch-injection → GUI-schema → i18n(en/fr) chain as every other `led_*` parameter.

## Test plan
- [ ] `cd ros2 && make build-pkg PKG=mowgli_leds && PACKAGES="mowgli_leds" ./scripts/test.sh` — 4 new `test_led_pattern.cpp` cases: evenly-spaced count lights exactly the requested pixels, explicit IDs take priority over count, an out-of-range ID is dropped rather than crashing, and a count larger than `led_count` clamps to the whole ring instead of an out-of-bounds write.
- [ ] `python3 ros2/scripts/check_config_drift.py` (already run locally, clean)
- [ ] `cd gui/web && npx tsc --noEmit && yarn lint && yarn test` (schema/template defaults kept in lockstep for the generic Go parity test; i18n en/fr key parity verified locally; `LedsSection.test.tsx` updated — note the existing charge-complete spinbutton test was switched from `length-2`/`length-1` indexing to fixed indices 10/11, since the new fields are appended after it)
- [ ] `cd gui && go test ./pkg/api/...` (`settings_leds_test.go` expected-key list updated)


cedbossneo (commented):
Reviewed against `dev`: plumbing is complete and parity holds (template 0 / 0.15 / `""` ↔ launch ↔ node clamp ↔ schema ↔ Go parity ↔ GUI section/claim list ↔ en/fr ↔ README/codemap/parameters.md), defaults leave #579 byte-identical, and the four new `test_led_pattern.cpp` cases do exercise the new render branch (even spacing {0,4,8,12}, ids beat count, out-of-range dropped, count 999 clamps). CI green. No safety surface.

One small fix before merge:

- **`led_ring_node.cpp:50-79` — `ParseIndicatorIds` is untested.** It is the only new input-validation path (free-text from the GUI) and sits in an anonymous namespace of the rclcpp TU, so neither gtest can reach it (`CMakeLists.txt:103-108` are header-only targets). Please move it into a header (`led_pattern.hpp` or a small `indicator_ids.hpp`) and add two cases: well-formed with spaces, and malformed / negative / overflow tokens skipped.

Optional in the same push:
- `led_ring_node.cpp:177`: IDs ≥ `led_count_` are silently dropped at startup; if every ID is stale the ring goes fully dim and the count is not used. A one-time `RCLCPP_WARN` listing the dropped IDs would save a "my indicator vanished" report.
- `LedsSection.test.tsx:139-142`: indices 10–13 are correct, but the comment forgets the Behavior card's five fields also precede the charge-complete pair.

Will merge as soon as the parser test is in.

cedbossneo:
Pushed the small fix to your branch directly (9d909090) so this can go in without another round-trip: `ParseIndicatorIds` moved to `include/mowgli_leds/indicator_ids.hpp`, three gtests in `test/test_indicator_ids.cpp`, a one-time startup warning for IDs past the end of the strip, the test comment, and the codemap rows. `mowgli_leds` builds and passes 75 tests in the dev image. Merging once CI is green — shout if you'd rather rework any of it.

cedbossneo (approved):
Reviewed: plumbing and parity complete, defaults keep #579 byte-identical, parser now unit-tested; CI green, merging.

PR #597, opened 2026-09-13
fix(mowgli_map): reject dock calibration when fused yaw disagrees with COG heading

### LINKED ISSUE #446: Docking calibration fails

fwelvering:
Fixes #446

## Root cause

`set_docking_point`'s existing yaw-convergence gate only proves the fused yaw was **STABLE** while a GPS-averaged dock position is captured -- never that it was **ACCURATE**. `/gps/pose_cov`'s lever-arm correction (in `navsat_to_absolute_pose_node`) rotates the antenna offset by whatever the fused yaw currently is, so a stable-but-biased yaw (e.g. a settled-wrong magnetometer lock) silently corrupts the averaged dock position by `lever_arm_length * sin(yaw_bias)` without ever tripping the existing gate.

Issue #446's reported ~10 cm lateral dock-position error is consistent with exactly this mechanism (~19 degrees of bias at the default 0.3 m forward GPS offset).

## Fix

Adds gate (2b) to `on_set_docking_point`'s GPS-position-averaging path: cross-check the fused yaw against `/imu/cog_heading`, an independent, GPS-motion-derived heading (`cog_to_imu_node`) that never depends on magnetometer/gyro lock. The gate fails closed, matching every other gate in this handler:

- Rejects when no sufficiently fresh/confident COG sample is available (`dock_set_cog_max_age_s`, `dock_set_cog_max_sigma_rad`).
- Rejects when the fused yaw disagrees with the COG heading by more than `dock_set_yaw_bias_max_rad`.

The gate is scoped to the GPS-position-averaging path (`use_gps_position=true`) only -- a manual map-drag / settings edit (`use_gps_position=false`) is unaffected, since the operator is explicitly setting the position by hand in that case.

The new thresholds follow the same maintainer-only, ROS2-param-only convention as their sibling `dock_set_*` thresholds (`dock_set_gps_accuracy_max_m`, etc.) -- not template/GUI-exposed.

## Testing

- New `DockCalibrationGateTest` suite in `test_map_server.cpp` (7 tests): rejects on missing/stale/uncertain COG sample, rejects on yaw disagreement, accepts across the angle-wrap boundary, accepts and correctly averages GPS when yaws agree, and confirms the manual-set path is unaffected.
- Full `mowgli_map` package test suite passes (175 tests, 0 failures).
- clang-format clean.


cedbossneo (changes requested):
Thanks for digging into the dock-calibration lever-arm bias — the diagnosis (the averaged antenna position is de-rotated with a yaw that may be wrong, leaving `0.3·sin(bias)` in the dock point) is right and worth fixing. The gate this PR adds cannot ship as written though, because it consumes a signal that does not exist at calibration time and compares it against the very quantity being recalibrated. Requesting changes.

### CRITICAL — the COG signal is a stationary latch on the dock, and the sigma gate rejects every real calibration

`cog_to_imu_node` rejects every fix while stationary; what arrives on `/imu/cog_heading` on the dock is the **stationary latch** republished at `stationary_seed_rate_hz` (2 Hz), with variance inflated by `stationary_yaw_drift_rate` (0.005 rad/s) × time since the last motion sample (`cog_to_imu_node.cpp:159-161`). Two consequences:

- `map_server_node.cpp:328` stamps `last_cog_yaw_time_ = now()` at receipt, so the "age" is the republish cadence (~0.5 s) and `dock_set_cog_max_age_s` (30 s) is dead code while the latch lives. The only operative check is sigma.
- One COG sample at the 0.10 m `min_baseline_displacement_m` has `sigma_pos_yaw = atan2(2·hypot(σ,σ), 0.10)`. The reference UM980 reports 14–20 mm under Fixed, giving 0.38–0.5 rad (22–30°) per sample — already above the 0.15 rad threshold **before** any drift. Even a 3 mm receiver (base 0.085 rad) closes the gate ~24 s after the last motion, and `calibrate_imu_yaw_node` spends that window on purpose (8 s charger wait, then up to 8 × 5 s retries waiting for gate (3) to converge, `calibrate_imu_yaw_node.cpp:182-186, 1624-1650`). Gate (3) needs quiet yaw; gate (2b) needs a fresh motion sample; the flow cannot satisfy both.

### CRITICAL — the "fused yaw" is the OLD dock yaw, so the gate locks out the only repair path

While `is_charging`, fusion_graph force-anchors every node onto the FULL `dock_pose`, yaw included (`fusion_graph_node_callbacks_a.cpp:358-372`). The bias the gate measures is therefore `old_dock_yaw − COG`, not a settled-wrong magnetometer lock. When the stored yaw is wrong by more than 8.6°, the gate rejects the MOTION-yaw call — the one non-circular way to fix it — and the log's "drive forward and retry" cannot help because the yaw re-pins the instant charging resumes. Permanent lockout for exactly the case the PR targets. Issue #446's reporters, meanwhile, are stuck on gate (3) ("std 1.402 > 0.500"), which this PR leaves untouched while adding a stricter gate on top.

### Suggested direction

Keep the diagnosis, drop the gate, and **correct** instead of reject: in MOTION mode re-project the averaged antenna position with `req->yaw_rad` (undo `R(yaw_tf)·L`, apply `R(yaw_rad)·L`, lever arm from TF `base_footprint → gps_link`), or average raw `/gps/fix` through `wgs84_projection.hpp`. Add a test with COG sigma 0.3 rad and a stored yaw 15° off that must SUCCEED and land within 1 cm — the current seven tests inject σ = 0.01 rad (0.57°, ~40× tighter than the node can produce at a 0.10 m baseline), age 0, and never exercise the MOTION path or the 2026-09-02 real case (−48.8° → −53.55°, a legitimate 5° correction).

### Smaller items

- HIGH `map_server_node.cpp:326-328`: if any freshness check survives, use `header.stamp`, not receipt time.
- LOW `area_manager.cpp:847`: `recent_yaws_.back()` relies on gate (3); with `yaw_convergence_min_samples` set to 0 at runtime it is `back()` on an empty deque.
- LOW: the new thresholds are cached at construction while the sibling `dock_set_*` are read live; `docs/claude/parameters.md` is not updated; `calibrate_imu_yaw_node.cpp:1174`'s error string does not mention the new gate.
- QoS is correct (`SensorDataQoS`), CI is green, the wrap-around test is a good addition.

cedbossneo (approved):

Issue #497, opened 2026-08-24
Headland pass overlap


fwelvering:
### Component

ROS2 Stack

### Problem or Motivation

It would be great if there was an configuration item to tell the mower not to cros more than X headland passes.

I configured it to mow 3 headland passes. (which it not uses, never does more than 1)
IT would be perfect if it did only cros in my case 2 headland passes so it stays clear from objects and within the boundry 

### Proposed Solution

Make it configurable and implement headland pass setting

### Would you work on this?

I can help test

cedbossneo:
Cross-referencing **#499**, which measured something that may explain the *"which it not uses, never does more than 1"* part of this report.

While investigating violent swath-end turns that dig the lawn, a connector-outcome counter was added to the coverage planner (#502). On its first run it showed that on a hole-free convex field, **only 1 of 32 swath joins gets a real turn-around arc** — the other 31 fall through to a straight "blind" connector.

The cause is the headland apron, not the turn radius:

| apron (`num_headland_passes` x `op_width`) | attempted joins | arc | straight fallback |
|---|---:|---:|---:|
| shipped (2 passes -> 0.32 m) | 32 | **1** | **31** |
| wide (8 passes -> 1.28 m) | 26 | 24 | 2 |
| swath spacing >= 2R | 10 | 10 | 0 |

At swath spacing below twice the turn radius, the turn-around must be an omega loop needing `sqrt(4R^2 - (R + d/2)^2) + R` = **0.43 m** of forward extent, against an apron of only **0.32 m** at the shipped defaults. Nothing in the configured radius range fits, so the arc search fails and the planner joins the swaths with a straight segment instead — which leaves a sharp corner the filleting step cannot round (measured: 31 corners over 88 degrees, maximum exactly 90.0, minimum implied radius 0.018 m).

That is a plausible mechanism for headland passes not being used the way they are configured: the apron is what the turn-around has to fit inside, and today it mostly does not.

Two things worth separating in this request:

1. **The bug** — passes configured but not used as expected. Likely the same area as #499; `num_headland_passes` is now the identified lever there and its effect is measurable via the new counter (expect a WARN on every plan reporting the fallback share).
2. **The enhancement you asked for** — a limit on how many headland passes the mower may *cross* during a connector, to stay clear of objects and inside the boundary. That is genuinely new and not covered by #499.

[...]

Issue #593, opened 2026-09-13
coverage: restore a real anti-fragmentation bound and pivot in place for tiny sub-path gaps


cedbossneo:
Since `40d0c30b` every sub-path boundary is a blade-off Nav2 transit and the coverage tests' fragmentation bounds were replaced by `subs.size() <= rings + swaths` (an empty ceiling). Field 2026-09-12: a 0.40 m boundary transit spun for 164 s (fixed by the transit goal checker + watchdog in `dffc2420`, but the transit itself is still pointless at that distance). Add a test bounding sub-path count on the shipped geometry (5 headland passes, 0.20 m radii), warn/refuse in `planCoverage` above a split ratio, and pivot in place (blade off until PRE_ROTATE ends) instead of a Nav2 transit when the gap is below `kSegmentTransitGapM`. Review finding H3 (motion).

Opened from the 2026-09-10 branch review of #565 (feat/lidar-map-anchor), merged with the anchor OFF by default. https://claude.ai/code/session_011wKpGhX81i85jF71nhidSm

cedbossneo:
Closed by #719 (merged just now), on top of #716.

The motion side was already solved by #716: planner-certified short pivot joins plus FTC's `PIVOT` state, which turned a 135-sub-path plan into 10 on the 2026-09-21 lawn. #719 adds the missing verification: a non-vacuous recorded-area regression using shipped settings bounds both the planner diagnostics and the executor-facing result (at most 4 splits, at most a 20 % split share, `subpaths == splits + 1`, at most 5 sub-paths), and single-sources the existing 25 % connector-split warning threshold.

No motion or blade behaviour changed. As #719 notes, #716's blade-on pivot motion itself still needs field validation.

Closing manually — PRs here target `dev`, so the closing keyword doesn't fire.

Issue #388, opened 2026-07-23
coverage: outermost-ring connector start falls outside the clearance ring → seg-1 transit unreachable → skipped


cedbossneo:
## Summary
Coverage planning emits the **outermost-ring connector outside the clearance ring**, placing the first swath/ring **start pose outside the allowed area**. The blade-off transit to that start is then un-drivable (RPP forward-collision refuses to move), so the whole sub-path is **skipped**. This is the primary reason area coverage collapses to the outer rings only.

## Evidence (field run 2026-07-23, `recorded_area_1`; full log preserved on robot `/home/ubuntu/mow_20260723_run.log`)
- Every plan logs, e.g.:
  - `PlanCoverage: 4/15361 continuous-path poses outside the outermost-ring clearance ring — connector geometry bug (blade may approach the boundary)`
  - (area 0) `3/13118 continuous-path poses outside the outermost-ring clearance ring`
- At the seg-1 start the robot is reported **outside the field**:
  - `BOUNDARY VIOLATION: robot at (0.23,12.97) — 0.03m outside nearest allowed area` … climbing to `0.12m outside`
- Result: `RegulatedPurePursuitController detected collision ahead!` (continuous ~34 s) → `Controller patience exceeded` (error_code 104, ×8) → `FollowStrip: segment 1 transit failed — skipping`.

## Suspected root cause
The continuous-path connector arcs (`buildContinuousPath` / `buildContinuousSubPaths`) for the outermost ring are not bounded to the chassis-safety inset — a few poses fall outside the clearance ring / area boundary. Combined with `local_costmap.inflation_radius=0.58` the start neighborhood is high-cost, so RPP cannot reach it. (Note: the ERROR is already detected and logged but not corrected.)

## Relevant files
- `ros2/src/mowgli_coverage/` (coverage server: `planBoustrophedon`, `buildContinuousPath`/`buildContinuousSubPaths`, the clearance-ring check that emits this ERROR)
- `ros2/src/mowgli_behavior/src/coverage_nodes.cpp` (`FollowStrip` transit / detour skip path)
- Context: `OBSTACLE_WEDGE_RECOVERY_SPEC.md` (Part A — prevention), CLAUDE.md invariant 7

## Acceptance criteria
- No `poses outside the outermost-ring clearance ring` ERROR for `recorded_area_1` (and the sim polygon) — the connector geometry stays inside the safety inset.
- Seg-1 transit reaches the ring start; area coverage no longer collapses to outer rings.
[...]

PR #598, opened 2026-09-14
feat(mowgli_coverage): limit how many headland passes a turn-around may cross

fwelvering:
Implements "option 2" from #497, as requested by the reporter.

## Problem

`set_docking_point`'s yaw gate is unrelated -- this is a separate issue on the coverage planner. Turn-around connectors between swaths are bounded by `connector_clearance_boundary`, which always sits on the **outermost** headland ring's centerline regardless of how many rings are configured. So with e.g. `num_headland_passes: 3`, a U-turn can still swing all the way out through every ring toward the recorded boundary, putting the chassis nose close to it and to nearby objects.

## Fix

Adds `connector_max_headland_passes`: the number of headland passes, counted from the mainland edge outward, a turn-around connector may cross.

- `0` (default) keeps today's behavior -- unlimited, bounded to ring 0's centerline.
- A value below `num_headland_passes` moves the clearance boundary inward to ring `(num_headland_passes - value)`'s centerline instead, leaving the outermost rings a no-turn zone.
- No effect with headland rings disabled (`num_headland_passes < 0`).

Example: 3 headland passes configured, limit set to 2 -- turns stay off the outermost ring, one ring-width further from the boundary than before.

Wired end to end:
- `coverage_server` param (read live, like the sibling `connector_turn_radius` knob)
- `navigation.launch.py` injection
- `mowgli_robot.yaml` template default
- GUI Mowing settings (a Select tied to the configured Headland Passes count, disabled when that count is Auto/None)
- `paramCatalog` / i18n (en, fr)

## Testing

- New `coverage_planning` gtests verify the clearance-ring erosion depth for the default, clamped, and rings-disabled cases against the real Fields2Cover v3 library (built locally from source to validate, since the devcontainer image only ships F2C v2) -- full suite: 234 tests, 0 failures.
- New `MowingSection` GUI tests (default/options/disabled states) -- 10/10 passing.
- New `test_launch_injection.py` guard confirms the wiring -- 10/10 passing.
- `TestSchemaDefaultsMatchTemplate` (Go) passes -- schema default and template default agree.
- `go build` / `go vet` clean.
- clang-format clean.


cedbossneo (changes requested):
Thanks for taking #497 on — the plumbing (template → launch → node → schema → i18n → docs, plus the injection test) is complete and CI is green. The core geometry has a problem though, so requesting changes.

### CRITICAL — tightening `connector_clearance_boundary` collapses the outer rings

`connector_clearance_boundary` is not only the arc-fit bound. `coverage_server.cpp:594/614` passes it as `boundary` into `buildContinuousSubPaths`, where it is (a) the `allInside` bound for every Dubins join, (b) the `allInside` test on straight fallbacks, and (c) the #388 per-pose clamp applied to EVERY pose of EVERY sub-path, rings included (`coverage_planning.cpp:1702-1708`, `clampInsideRing` projects to 2 cm inside the ring).

With N=3 / P=2 (the reporter's case) ring 0 lies one `op_width` outside the new boundary, so every ring-0 pose is projected onto ring 1's centerline: the outermost pass is never driven, ring 1 is driven twice, the perimeter loses a full `op_width` band. Same with N=5 / P=4. It is silent — the out-of-bounds verify runs on the clamped path and `result->segments` still shows ring 0 in the GUI. Ring 0's exit pose is also outside the boundary, so the ring-0 → ring-1 join fails `allInside` and becomes a blade-off transit.

The limit needs a **separate turn-envelope polygon** used only inside `buildConnector`'s `allInside` for swath-end joins; the clamp, the verify, the ring joins and `HeadlandRingStaysInsideClearance` must keep ring 0.

### HIGH — at every value that actually restricts, each U-turn becomes a transit

At R=0.20 / spacing 0.16 the shortest Dubins word (RLR) needs ~0.49 m beyond the swath ends. Apron available is (P−0.5)·op_width: P=4 → 0.56 m fits, P≤3 does not. Since 40d0c30b a straight fallback with a 180° discontinuity is a SPLIT (`straightFallbackIsContinuous`), so P≤3 of 5 — or any P with N=3 — turns every swath join into a ~0.16 m blade-off Nav2 transit, exactly the sub-metre transit problem #593 is open about. The hpp/yaml comments ("that fallback still passes allInside … the sub-path does NOT fragment") predate 40d0c30b and are no longer true. Please either refuse/clamp server-side when the remaining apron cannot fit the omega turn, or at minimum add a fragmentation test on the shipped geometry.

Note for #497 itself: with 3 passes and R=0.20 the apron is 0.40 m < 0.49 m, so no arc fits even unlimited — the reporter's field fragments on `dev` today and this knob cannot help them; more passes (or #593's pivot-in-place) is the fix there.

### HIGH — `MowingSection.tsx:31`

`HEADLAND_PASSES_TEMPLATE_DEFAULT = 2` while template and schema say 5 (pre-existing drift, but the new Select is built on it): on a sparse robot the operator is offered 1..2 while the server runs 5 rings; picking 2 gives P=2 of 5 → apron 0.24 m → every U-turn a transit plus the ring collapse above. Please derive it from the schema default.

### MEDIUM / tests

- `coverage_server.cpp:607-615`: the INFO log gates on `num_headland_passes_ > limit`; in AUTO (0) the limit still applies to the derived ring count with no log (the GUI disables the Select for Auto, yaml does not).
- Neither new gtest calls `buildContinuousSubPaths` with a limit, so neither sees the collapse, the ring-join splits, or the sub-path count. Please add one on the shipped geometry (5 rings, op 0.16, R 0.20) asserting ring-0 poses are unchanged with a limit set and asserting the sub-path count for P=4.

Happy to re-review once the envelope is split out.

cedbossneo (dismissed):
Thanks for taking #497 on — the plumbing (template → launch → node → schema → i18n → docs, plus the injection test) is complete and CI is green. The core geometry has a problem though, so requesting changes.

### CRITICAL — tightening `connector_clearance_boundary` collapses the outer rings

`connector_clearance_boundary` is not only the arc-fit bound. `coverage_server.cpp:594/614` passes it as `boundary` into `buildContinuousSubPaths`, where it is (a) the `allInside` bound for every Dubins join, (b) the `allInside` test on straight fallbacks, and (c) the #388 per-pose clamp applied to EVERY pose of EVERY sub-path, rings included (`coverage_planning.cpp:1702-1708`, `clampInsideRing` projects to 2 cm inside the ring).

With N=3 / P=2 (the reporter's case) ring 0 lies one `op_width` outside the new boundary, so every ring-0 pose is projected onto ring 1's centerline: the outermost pass is never driven, ring 1 is driven twice, the perimeter loses a full `op_width` band. Same with N=5 / P=4. It is silent — the out-of-bounds verify runs on the clamped path and `result->segments` still shows ring 0 in the GUI. Ring 0's exit pose is also outside the boundary, so the ring-0 → ring-1 join fails `allInside` and becomes a blade-off transit.

The limit needs a **separate turn-envelope polygon** used only inside `buildConnector`'s `allInside` for swath-end joins; the clamp, the verify, the ring joins and `HeadlandRingStaysInsideClearance` must keep ring 0.

### HIGH — at every value that actually restricts, each U-turn becomes a transit

At R=0.20 / spacing 0.16 the shortest Dubins word (RLR) needs ~0.49 m beyond the swath ends. Apron available is (P−0.5)·op_width: P=4 → 0.56 m fits, P≤3 does not. Since 40d0c30b a straight fallback with a 180° discontinuity is a SPLIT (`straightFallbackIsContinuous`), so P≤3 of 5 — or any P with N=3 — turns every swath join into a ~0.16 m blade-off Nav2 transit, exactly the sub-metre transit problem #593 is open about. The hpp/yaml comments ("that fallback still passes allInside … the sub-path does NOT fragment") predate 40d0c30b and are no longer true. Please either refuse/clamp server-side when the remaining apron cannot fit the omega turn, or at minimum add a fragmentation test on the shipped geometry.

Note for #497 itself: with 3 passes and R=0.20 the apron is 0.40 m < 0.49 m, so no arc fits even unlimited — the reporter's field fragments on `dev` today and this knob cannot help them; more passes (or #593's pivot-in-place) is the fix there.

### HIGH — `MowingSection.tsx:31`

`HEADLAND_PASSES_TEMPLATE_DEFAULT = 2` while template and schema say 5 (pre-existing drift, but the new Select is built on it): on a sparse robot the operator is offered 1..2 while the server runs 5 rings; picking 2 gives P=2 of 5 → apron 0.24 m → every U-turn a transit plus the ring collapse above. Please derive it from the schema default.

### MEDIUM / tests

- `coverage_server.cpp:607-615`: the INFO log gates on `num_headland_passes_ > limit`; in AUTO (0) the limit still applies to the derived ring count with no log (the GUI disables the Select for Auto, yaml does not).
- Neither new gtest calls `buildContinuousSubPaths` with a limit, so neither sees the collapse, the ring-join splits, or the sub-path count. Please add one on the shipped geometry (5 rings, op 0.16, R 0.20) asserting ring-0 poses are unchanged with a limit set and asserting the sub-path count for P=4.

Happy to re-review once the envelope is split out.

fwelvering:
Thanks for the detailed review — pushed a follow-up (71f14325, plus two format-only fixups 5b610ce1/16121104) addressing all of it:

CRITICAL — ring-0 collapse. Split the boundary in two: connector_clearance_boundary now always stays ring 0's centerline, unaffected by connector_max_headland_passes. A new BoustrophedonPlan::swath_turn_envelope carries the tighter ring(n_rings − limit) bound, and buildContinuousSubPaths takes it as an optional swath_turn_boundary argument used only for buildConnector's allInside test on joins between two mainland swath segments. Ring-to-ring joins, the ring-to-first-swath transition, the #388 out-of-bounds clamp, the whole-path corner-fillet pass, and the server's verify all stay bound to ring 0 — so the limit can no longer clamp ring 0's own poses onto a different ring's line. New test SwathTurnEnvelopeNeverRelocatesRingZero exercises buildContinuousSubPaths with a limit and asserts ring 0's vertices are actually driven, not relocated.

HIGH — apron fragmentation. Added two gtests on the shipped geometry (op 0.16, R 0.20) pinning both sides of your own math: ShippedGeometryTightSwathLimitFragmentsUTurns (limit=3/5, apron 0.40 m) asserts strictly more splits than the unlimited baseline; ShippedGeometryP4OfFiveKeepsTurnAroundArcs (limit=4/5, apron 0.56 m) asserts no extra fragmentation. Both compare against a same-plan baseline rather than hardcoding a pose/sub-path count, since I couldn't run colcon locally to pin exact numbers (see note below).

HIGH — MowingSection.tsx:31. HEADLAND_PASSES_TEMPLATE_DEFAULT is gone; the sparse-robot fallback now reads defaults.num_headland_passes (the loaded schema default) via a new defaults prop threaded from SettingsPage.tsx, falling back to a literal only before that map has loaded — and that literal is now 5, matching the schema, not the stale 2.

MEDIUM — AUTO log gating. The "turn-around connectors limited" INFO log now gates on the plan's actual resolved ring count (new BoustrophedonPlan::n_headland_passes) instead of the raw num_headland_passes_ config, so it fires in AUTO mode too.

MEDIUM — stale doc comment. Fixed the hpp comment claiming a starved-apron straight fallback "does NOT fragment" — no longer true since 40d0c30b's straightFallbackIsContinuous; updated to match the tested behavior.

cedbossneo (approved):

PR #601, opened 2026-09-14
fix: make DIG_OBSTRUCTION recoverable by Play (after lifting clear) or Home

cedbossneo:
## Problem

A robot held in **DIG_OBSTRUCTION** (repeat-dig escalation, Invariant 16) could not be recovered by the operator (field report 2026-09-14):

- **"Moved it by hand, Play did nothing."** The bridge cleared `dig_escalated` only when the robot reached the charger. A robot lifted onto open grass stayed latched, so the tree's `DigObstructionGuard` refused START forever.
- **"Pressed Home, it never moved."** Up to three 0.60 m PENDING dig keepouts get stamped around the robot; with the v1.2.0 centred square the dock transit planned from inside them (START_OCCUPIED). On today's heading-biased keepout HOME did work in a live escalation (see below), but nothing guaranteed it.
- The GUI did not know the state name at all: fallback label, no state-graph node, and the map toolbar offered **Pause** instead of Play.

The hold itself is correct and unchanged: it publishes numeric state 1 (IDLE), which the firmware turns into a wheel + blade hard stop for as long as a wedged robot is held.

## Fix

1. **Displacement clear (bridge).** `DigEscalationClearedByDisplacement`: the latch is released once the fused pose is `kDigEscalationClearFactor` (2×) `dig_escalate_radius_m` = 1.0 m from the escalation point — carried clear, or driven out. Shuffling inside the hole stays latched. Charger clear unchanged.
2. **HOME drops the keepouts under the robot (map_server + BT).** New `~/discard_dig_keepouts_near_robot` (`std_srvs/Trigger`) removes PENDING `SOURCE_DIG` proposals that contain / lie within 0.60 m of the robot; accepted keepouts and farther proposals stay. `HomeSequence` calls it through `DiscardNearbyDigKeepouts` under `ForceSuccess(Sequence(IsDigEscalated, …))`, after `RETURNING_HOME` and before `DockRobot`. Bounded 3 s ack wait, always SUCCESS.
3. **GUI.** `DIG_OBSTRUCTION` dashboard card (danger tone, recovery hint en/fr), state-graph alias, toolbars treat it as idle (Continue + Home).

## Tests

- `test_dig_escalation.cpp`: 4 new cases (inside hole no clear, carried clear clears, 2× boundary, disabled radius).
- `test_map_server.cpp`: 2 new cases (drops only the proposal under the robot; keeps accepted keepouts and far proposals).
- `test_dig_obstruction_recovery.cpp` (new, structural like `test_dock_motion_gate`): HOME gates the discard on `IsDigEscalated` before `DockRobot`; the hold still publishes state 1.
- GUI: `BTStateGraph.test.tsx` state list extended; vitest 417/417, tsc clean.
- Local docker (dev image): mowgli_map 49/49, mowgli_behavior gtests all green (only the pre-existing cpplint ctest entry fails, on untouched files), mowgli_hardware `test_dig_escalation` green.

## Live evidence (2026-09-14, dev image eec1daa5)

Genuine escalation at 12:54 against the boundary hedge: three latches in 13 s (encoders 0.46–0.55 m, raw RTK 0.00–0.01 m, GNSS σ 0.014 m). HOME from the GUI → `RETURNING_HOME` (state 2, firmware unlocked) → staging in 85 s → dock contact → latch cleared on the charger. So the HOME path works on the heading-biased keepout; this PR makes it robust when the proposals do cover the footprint, and adds the missing Play-after-lift exit.

**Safety-relevant:** changes when the escalation latch releases (physical behaviour: Play becomes accepted 1 m away from the dig point). It never releases while the robot is still at the spot, and the firmware stays the sole blade authority.

https://claude.ai/code/session_011wKpGhX81i85jF71nhidSm


fwelvering:
https://github.com/mowglinext/mowglinext/pull/600 <- same issue/fix

PR #570, opened 2026-09-08
fix(gnss): a repeated observation sequence under a newer stamp is a cached publication, not invalid provenance

cedbossneo:
## What
`gnss_observation_freshness.hpp`: a repeated `position_observation_sequence` under a newer header stamp is a **cached publication**, not invalid provenance. It neither refreshes the deadline nor invalidates GNSS authority; a frozen receiver still fails closed at `maximum_age` with no new sequence.

## Why
Field 2026-09-08: `LocalizationGuard` paused mowing for 2 s every few minutes ("GNSS feed stale" + "GNSS receipt stamp is zero/future; authority denied" from both `behavior_tree_node` and `hardware_bridge` — 1260 times in 30 h, 205 pauses in one hour while docked). Each pause interrupts `FollowStrip` and consumes one of the five dispatch attempts; the mow gave up after three minutes with the robot on the lawn.

Caught live: at 09:25:48 `/gps/status` repeated sequence 1827777 under a newer stamp, and at that millisecond both consumers denied authority. The universal_gnss receiver stamps its status with the publication time and uses the observation **count** as the sequence, so a status published twice between two observations legitimately repeats the sequence.

## Test plan
- [x] `test_gnss_observation_freshness` 6/6 locally (new `RepeatedSequenceUnderNewerStampIsCachedNotInvalid`)
- [ ] CI ROS2 build & test
- [ ] Robot: no more "zero/future" warnings / "GNSS feed stale" pauses on the dock over an hour

https://claude.ai/code/session_01D46c8dsyRG7k8Q7Djs1NjZ

PR #625, opened 2026-09-15
feat(gnss): integrate Universal GNSS sidecar

Pepeuch:
## Summary

Integrate Universal GNSS as the canonical external GNSS runtime for MowgliNext, using the official Universal GNSS sidecar and a lightweight contract bridge inside `mowgli-ros2`.

This PR completes the GNSS runtime split, preserves MAVROS/GNSS independence, adds safe GUI configuration/recovery behavior, and updates CI / deployment handling for the new architecture.

Hardware validation was performed with a u-blox F9P and with a Unicore UM982 setup.

Known follow-up outside this PR:
Universal GNSS RC2 still needs a u-blox post-apply active baud verification fix (`active_verified_baud` / `current_baud_verified`). MowgliNext now safely restores the GNSS service even when that condition causes a partial apply result.

Physical mower motion, blade control, e-stop and safety gates are unaffected.

## Changes

- Move the full Universal GNSS runtime out of `mowgli-ros2`.
- Run Universal GNSS exclusively through the official external `mowgli-gps` sidecar.
- Pin Universal GNSS RC2 interface/runtime contract.
- Keep only the lightweight `universal_gnss_msgs` interface dependency inside the ROS2 image.
- Add / update the Mowgli GNSS contract bridge:
  - `/universal_gnss_receiver/status` → `/gps/status`
  - `/universal_gnss_receiver/rtcm` → `/rtcm`
  - `/gps/fix` remains published directly by Universal GNSS.
- Launch `mowgli_gnss_bridge/universal_gnss_topic_bridge` automatically from `mowgli-ros2` when `GNSS_STACK=universal`.
- Keep `GNSS_STACK` independent from `HARDWARE_BACKEND`, including:
  - `HARDWARE_BACKEND=mavros`
  - `MAVROS_ENABLED=true`
  - `GNSS_STACK=universal`
- Keep Universal GNSS as the sole NTRIP owner.
- Add targeted serial-device/GID handling for temporary GNSS configuration containers.
- Add GNSS baud detection / configuration support with explicit current, target and verified baud semantics.
- Add GNSS GUI plan/apply/restart flows.
- Ensure a GNSS service stopped by GUI Apply is always reconciled afterward, including:
  - detection failure
  - apply failure
  - partial result
  - request cancellation
  - persistence failure
- Detach GNSS restore cleanup from HTTP request cancellation while keeping it timeout-bounded.
- Prevent stale Unicore-only options from leaking into u-blox/NMEA configuration:
  - receiver model
  - signal group
  - rover dynamic mode
- Preserve Unicore expert settings while receiver family is `auto`.
- Add/update installer and compose generation for the Universal GNSS sidecar.
- Track the external Universal GNSS image in the managed updater/deployment metadata.
- Update ROS2 Docker smoke tests for the interface-only Universal GNSS dependency.
- Fix YAML-first NTRIP fallback behavior.
- Add GNSS launch/config/apply regression coverage.
- Fix ROS2 launch tests for the Lyrical `launch_ros.actions.Node` API.
- Rename the ROS2 CI required check from Kilted to Lyrical.
- Stabilize an unrelated IrriSense async test discovered while validating the branch; runtime IrriSense behavior is unchanged.

## Component

- [x] ROS2 Stack (`ros2/`)
- [x] GUI (`gui/`)
- [x] Docker (`docker/`)
- [x] Sensors (`sensors/`)
- [ ] Firmware (`firmware/`)
- [x] Installer (`install/`)
- [ ] Documentation (`docs/`, `wiki/`)
- [x] CI/CD (`.github/`)

## Testing

- [x] Built successfully
- [x] Tested on hardware
- [ ] Tested in simulation
- [x] Unit tests pass
- [x] Manual testing

Hardware validation included:

- u-blox F9P:
  - Universal GNSS receiver detected and configured successfully.
  - `/universal_gnss_receiver/status` validated with live satellite/CN0/accuracy data.
  - `mowgli_gnss_bridge` validated against the live receiver.
  - `/gps/status` restored correctly with:
    - receiver vendor
    - fix / RTK state
    - horizontal accuracy
    - satellites used / visible
    - mean / max C/N0
  - GUI GNSS satellite and signal-quality graphs validated live.
  - NTRIP streaming validated through the Universal GNSS sidecar.

- Unicore UM982:
  - Universal GNSS sidecar validated with live receiver data.
  - NTRIP / RTCM flow validated.
  - Mowgli GUI / fusion graph integration validated.

Validation performed during the PR included:

- `go test ./...`
- GUI `vitest + tsc`
- GNSS-specific GUI regression tests
- ROS2 bringup / GNSS launch tests
- ROS2 formatting / static-analysis CI
- `git diff --check`
- compose validation, including managed-updater path: **47/47**
- amd64 / arm64 GUI Docker builds
- amd64 / arm64 ROS2 runtime Docker builds

## Checklist

- [ ] Documentation updated (if user-facing)
- [x] No unrelated runtime changes
- [x] Physical behavior (blades, motion, e-stop, safety gates) — unaffected
- [x] C++ touched → formatter CI passes
- [ ] `ros2/src/mowgli_interfaces/` touched → regenerated all three consumers (`firmware/scripts/sync_ros_lib.py`, `gui/generate_go_msgs.sh`, `gui/generate_ts_types.sh`) — not applicable
- [ ] COBS wire touched (`mowgli_protocol.h` / `ll_datatypes.hpp`) → bumped `MOWGLI_PROTOCOL_VERSION` and `kMowgliProtocolVersion` — not applicable
- [ ] `mowgli_robot.yaml` touched → `python3 ros2/scripts/check_config_drift.py` passes — not applicable


cedbossneo (commented):
Thanks for this — splitting the receiver/NTRIP runtime out of `mowgli-ros2` and keeping only the interface package plus a thin contract bridge inside it is the right shape, and the ROS-side contract is handled carefully. Review below; I have not run it on hardware, and I did not review the GUI plan/apply/restart flows in depth, so treat this as a review of the ROS2, launch, image and installer side.

## What is good

- **Invariant 1 is preserved.** `/gps/fix` is still published directly by the receiver node and never goes through the bridge, so `fusion_graph` keeps its raw `NavSatFix` input. The bridge only adapts `/universal_gnss_receiver/status → /gps/status` and `…/rtcm → /rtcm`.
- The `ros2/Dockerfile` change builds **only** `universal_gnss_msgs`, so the ROS2 image no longer carries the driver stack. That is a real reduction in surface.
- `deployment.json` pins the sidecar by **digest**.

## Blocking

### 1. Three environment-variable reads in `full_system.launch.py` (Invariant 15)

`dev` has no `os.environ` read in this launch file; this PR adds three:

```python
gnss_stack = os.environ.get("GNSS_STACK", "universal").strip().lower()
...
robot_params.get("gnss_receiver_family", os.environ.get("GNSS_RECEIVER_FAMILY", "auto"))
robot_params.get("gnss_frame_id",        os.environ.get("GNSS_FRAME_ID", "gps_link"))
```

- **`GNSS_STACK`** decides whether a ROS2 node exists in the graph, and it has no key in `mowgli_robot.yaml`, no template default and no GUI schema entry. `test_gnss_launch_config.py` (the `monkeypatch.setenv("GNSS_STACK", …)` cases) then pins that behaviour.
- **`GNSS_FRAME_ID`** is a live violation rather than a dead fallback: there is no `gnss_frame_id` key in the template, so the env var is the *effective* source of a ROS2 parameter default.
- **`GNSS_RECEIVER_FAMILY`** is dead today (the template already has `gnss_receiver_family: "auto"`), but it re-introduces the pattern.

This is exactly the `LIDAR_ENABLED` situation that was removed on 2026-08-31: as "a fallback when the yaml is silent" the env var silently overrode the operator's GUI choice and ran the stack in a coherent-looking half-state. The rule that came out of it is in `CLAUDE.md` (Invariant 15 and the matching "What NOT to Do" bullet): the ROS2 stack reads robot config through `robot_config_util.load_robot_params()` and nothing else.

Suggested shape: give `gnss_stack` the `resolve_lidar_enabled()` treatment (a `resolve_gnss_stack()` in `robot_config_util.py`, an absent-key default that matches the GUI schema default, a startup WARNING when the key is absent), add `gnss_frame_id` to the template, and drop the third fallback. The compose layer can still use `GNSS_STACK` to decide whether the **container** starts — that is the same split `LIDAR_ENABLED` has today — it just must not be what the launch file reads.

### 2. The installer pins the sidecar by a mutable tag

`deployment.json` has the digest, but the installer path does not use it:

```sh
# install/lib/config.sh
UNIVERSAL_GNSS_IMAGE_DEFAULT="ghcr.io/pepeuch/universal-gnss-ros2-lyrical:v0.1.4-rc2"
```

That image runs with `network_mode: host` and the GNSS serial device mapped in, and it comes from a personal namespace, so what a fresh install pulls should be immutable. Please pin the default by digest as well (`…@sha256:24ea6c1c…`), ideally sourced from the same place as `deployment.json` so the two cannot drift. Longer term it would be worth mirroring the image under the org namespace, but that is not a condition for this PR.

### 3. It conflicts with `dev`

GitHub reports `CONFLICTING`. #626 merged today and both PRs insert at the same anchor right after `robot_params = load_robot_params(...)` in `full_system.launch.py`; this PR additionally turns the `LaunchDescription([...])` tail into a `launch_entities` list. The two changes are semantically independent, so it is a short rebase, plus `install/mowglinext.sh` (a re-indent against an added `unset`).

## Please double-check on the rebase

- **The submodule re-pin** (`universal-gnss` `ab32f673 → c56356c2`). A branch cut before a re-pin silently reverts the gitlink when it merges, and `sensors-*` / `ros2-docker` have no PR trigger, so it lands red on `dev` with nothing flagging it. Worth an explicit `git submodule status` check after rebasing.
- **`/gps/status` QoS and freshness fields.** `position_observation_sequence` and `horizontal_accuracy_m` look copied through verbatim, which is what the #570 freshness guard and the dig detector's `DigTrustSigma` need — please keep a test on that, since a future bridge refactor that re-stamps or re-sequences the status would silently break both.

## Not reviewed

The GUI plan/apply/restart and recovery flows, the installer tests, and anything that needs a receiver on the bench. The known follow-up you list (RC2 post-apply baud verification on u-blox) sounds fine to leave out of this PR as long as the service restore you describe is covered by a test.

Pepeuch:
Thanks for the thorough review — this is very helpful.

I agree with the three blocking points.

1. I’ll remove the ROS-side environment fallbacks from `full_system.launch.py` and bring GNSS in line with Invariant 15, following the same pattern as LiDAR:
   - resolve the GNSS stack from robot config,
   - add the missing `gnss_frame_id` template default,
   - remove the `GNSS_*` fallbacks from the launch file,
   - update the tests so they no longer pin environment-driven ROS behavior.

2. I’ll pin the default Universal GNSS installer image by digest as well, matching the digest already present in `deployment.json`.

3. I’ll bring the branch back up to date with `dev` and resolve the launch/installer conflicts.

I’ll also explicitly verify the Universal GNSS submodule gitlink after the merge and keep the `/gps/status` freshness/provenance fields (`position_observation_sequence`, `horizontal_accuracy_m`) passed through unchanged.

The u-blox active-baud verification remains a Universal GNSS follow-up as noted in the PR; the MowgliNext service-restore path is covered separately.

Thanks again — this is exactly the kind of review that helps keep the integration clean.

PR #636, opened 2026-09-16
WIP / EXPERIMENTAL — STM32F401 USB DFU blocked on RM-MB ROM enumeration

liltaket:
## Status

> **WIP / EXPERIMENTAL — NOT HARDWARE VALIDATED**
>
> This PR will remain Draft. No STM32F401 mower/board HIL run, physical motor/blade inhibit measurement, real ROM enumeration, power-loss test, or ST-Link recovery exercise has yet been performed.

## Problem

Routine firmware updates for the supported STM32F401 board currently require opening the mower and attaching ST-Link. This adds a guarded, capability-driven USB update path through the STM32 factory ROM DFU bootloader while preserving the existing ST-Link/OpenOCD path as an explicit advanced and final-recovery option.

Initial USB support is deliberately limited to `BOARD_YARDFORCE500B` / `Yardforce500B` (`STM32F401VC`) and blessed prebuilt release artifacts. STM32F103 remains ST-Link-only. Existing F401 installations need one final ST-Link update before they can advertise the new USB-DFU capability.

## Architecture

1. The GUI explicitly starts one idempotent USB operation; custom builds always use ST-Link.
2. The backend validates schema-2 release metadata, board/environment/panel, MCU, address/flash bounds, exact size, vector table, protocol/version, and SHA-256 before DFU entry.
3. Fresh ROS status must prove compatible/capable firmware, IDLE ownership, no active motion/mow command, and healthy stopped blade telemetry.
4. The ROS bridge acquires firmware-update maintenance, clears queued/dig motion, and enforces zero motion, blade off, IDLE, and no emergency release at the final serial boundary.
5. Protocol v7 adds explicit `ENTER_DFU` plus a capability bit; F1 neither registers nor advertises it.
6. F401 firmware locks later inputs, clears targets/PID state, repeats drive/blade stop commands for a bounded 500 ms window, writes a magic+inverse one-shot token into an 8-byte linker-reserved `NOLOAD` SRAM section, then uses `NVIC_SystemReset()`.
7. The reset handler consumes and clears the marker before C runtime initialization. A fresh `SFTRSTF` is required; watchdog/power/brownout/low-power causes are rejected. `SFTRSTF|PINRSTF` is allowed because PINRSTF may accompany SYSRESETREQ; PIN-only remains rejected.
8. Hardware inhibit latches are loaded before GPIO output mode. A validated ROM vector then reaches a naked `MSR MSP` / `CPSIE i` / `BX` trampoline with no stack access after MSP changes.
9. Linux anchors `/dev/mowgli` to its physical USB port, accepts only one new `0483:df11` device on that port, and scopes every `dfu-util` call with VID/PID, port, and alternate setting.
10. The host writes `0x08000000`, uploads/read-compares the same byte range, uses DfuSe `:leave`, waits for CDC on the same port, then requires a newer completed firmware handshake with matching capability/protocol/version.

The ROM facts and transition design are grounded in ST [AN2606](https://www.st.com/resource/en/application_note/an2606-stm32microcontroller-system-memory-boot-mode-stmicroelectronics.pdf), [RM0368](https://www.st.com/resource/en/reference_manual/rm0368-stm32f401xbc-and-stm32f401xde-advanced-armbased-32bit-mcus-stmicroelectronics.pdf), and [AN3156](https://www.st.com/resource/en/application_note/an3156-usb-dfu-protocol-used-in-the-stm32-bootloader-stmicroelectronics.pdf). Host exit semantics use the documented [dfu-util DfuSe `:leave` modifier](https://dfu-util.sourceforge.net/dfu-util.1.html).

## Recovery model

- Normal USB mode rejects stale/false blade telemetry.
- Advanced **USB Recovery** may waive only that blade-telemetry check after a fresh physical stationary/blade-stopped confirmation.
- Recovery never waives firmware compatibility/capability, active motion/mow ownership, known IDLE state, artifact trust, physical USB identity, or the firmware make-safe sequence.
- If application USB cannot receive `ENTER_DFU`, identity becomes ambiguous, write/readback/return verification fails, or the outcome is uncertain, the operation ends `failed_requires_stlink` and maintenance stays latched.
- ST-Link remains unchanged and is the final recovery method.

### A normal update cannot begin while operation is commanded

Mechanisms: explicit user action; fresh host readiness; authoritative bridge service gate; firmware accepts DFU only in IDLE; no deferred start after a rejection.

Evidence: readiness/recovery tables, bridge policy tests, timeout/fault campaign, and firmware-native request-decision tests.

Residual risk: physical controller state and stop-frame receipt are HIL pending.

### Motion/blade commands cannot be re-enabled during the transition

Mechanisms: bridge final-boundary packet gate and zeroing; emergency-release suppression; stop latch after maintenance; firmware input lock, cleared targets/controllers, repeated zero/off frames; PE14/PE15 inhibit before ROM.

Evidence: pure ROS gate tests, firmware-native transition/tick-wrap tests, F1/F4 builds, and rebuilt F401 disassembly of the ROM trampoline.

Residual risk: actual pin levels, UART delivery, external controller behavior, and reset/ROM dwell require scoped HIL measurement.

### Wrong or corrupt firmware cannot reach erase/write

Mechanisms: strict manifest schema/metadata, exact target/MCU/address/size, vector validation, SHA-256, and pre-entry validation.

Evidence: focused manifest tests plus 262,144 deterministic artifact mutations (4,096 seeds × 64 mutations), all rejected.

Residual risk: release-signing/authenticity is still the existing HTTPS/GitHub release trust model; STM32 ROM DFU does not authenticate applications.

### Browser/API retries cannot silently start a second update

Mechanisms: one active backend operation, idempotency-key attachment, key persisted before POST, GET-only polling, detached operation context, and an atomic cancellation/irreversible-entry boundary.

Evidence: idempotency, reconnect, request-context, cancellation-race, and exclusive-operation tests.

Residual risk: operation state is in memory; backend restart loses history and requires deliberate operator reconciliation.

### Tool/process success cannot masquerade as update success

Mechanisms: device identity, scoped write, byte readback, leave, same-port CDC return, newer connection generation, completed compatible handshake, exact version/protocol/capability.

Evidence: subprocess argument/exit/readback tests, observer identity tests, transient-handshake regression tests, and dependency-boundary failures.

Residual risk: real enumeration timing and appliance permissions are HIL pending.

## Failure behavior

- Artifact/readiness failures stop before DFU entry and require a new explicit action.
- This first implementation performs no automatic retry.
- Cancellation is accepted only before an atomic irreversible boundary. Once `ENTER_DFU` may be dispatched, cancellation is refused.
- Any ambiguous request, enumeration, write, verification, leave, application-return, or runtime-handshake failure reports no success and retains the safety latch for ST-Link recovery.
- A later rejected operation cannot release a prior uncertain operation's maintenance ownership.
- Success alone releases maintenance, while the bridge retains a stop latch until a fresh IDLE high-level state is observed.

## Tests

Passed on macOS in the isolated `upstream/dev` worktree:

- `pio test -e native` — **8/8**
- `pio run -e Yardforce500` — PASS (F1 regression build)
- `pio run -e Yardforce500B` — PASS (F401 build)
- rebuilt ELF: `.dfu_magic` is 8 bytes at `0x2000fff8`; `_estack=0x2000fff8`; trampoline disassembly is exactly `msr MSP,r0`, `cpsie i`, `bx r1`
- `go test ./...` — PASS
- `go test -race ./pkg/providers` — PASS
- `npm test` — **60 files / 449 tests**
- `npx tsc --noEmit` — PASS
- changed frontend files lint — PASS with warnings only
- `npm run build` — PASS
- `board_defaults_parity.py` — PASS
- `test_package_release.py` — **1/1**
- `protocol_version_guard.py --check` — protocol v7, **18 structs / 18 IDs**, fingerprint matches
- `sync_ros_lib.py --check` — **16 headers** current
- Go and TypeScript message generators — no drift
- ROS C++ `clang-format` 18 dry check — PASS
- HIL helper syntax/help and refusal without exact bench-safety acknowledgement — PASS
- `git diff --check` — PASS

Local ROS Lyrical/vendor dependencies were not available, so `colcon build/test` is CI-pending. Full repository frontend lint currently reports 22 errors in unchanged files; all changed frontend files have zero lint errors.

## Deterministic stress / fault injection

- **262,144 artifact cases**: 4,096 seeds × 64 deterministic byte mutations; no corrupt artifact accepted.
- **262,144 production-engine boundary runs**: 4,096 seeds × 64 repetitions over 11 fixed dependency outcomes (artifact, readiness, CDC anchor, DFU request, DFU enumeration, write, readback, leave, application return, runtime verification, success). This is repeated boundary injection, not a claim of 64-event randomized sequences.
- Timeout injection at every bounded dependency; each invoked once, proving no automatic retry.
- Focused tests cover malformed/missing/inconsistent manifest metadata, wrong board/MCU/panel/environment/address/size/SHA/vector/protocol/version; expected/late/wrong/multiple/pre-existing DFU devices; same/wrong returning CDC port; dfu-util nonzero exit and partial readback; transient/stale/wrong/no runtime handshakes; false-RPM recovery; cancellation at the irreversible boundary; and maintenance ownership after an uncertain dispatch.

## Hardware testing still required

**No hardware validation has been performed.** Before this PR can leave Draft, execute and attach the full procedure in [`docs/STM32F4_USB_DFU.md`](docs/STM32F4_USB_DFU.md), including:

- one-time ST-Link install and full flash backup;
- scoped PE14/PE15, PA11/PA12, reset, and drive/blade UART evidence;
- real ROM enumeration and exact-port identity;
- upgrade/downgrade and 25 repeated successful cycles;
- wrong/corrupt artifact rejection before erase;
- active-operation rejection and false-RPM recovery;
- USB/process interruption at each stage;
- unrelated/multiple DFU-device rejection;
- incomplete-application ST-Link recovery and byte verification;
- separately supervised current-limited power-interruption work.

The included `tools/firmware/usb_dfu_hil.py` runner requires an exact blade-removed/wheels-constrained/ST-Link-ready acknowledgement, uses one POST per cycle followed only by GET polling, appends JSONL evidence, and stops at the first non-success. It does not control motion, blade, power, USB unplugging, or ST-Link.

## Remaining risks

- Physical drive/blade inhibition, external-controller receipt, USB clock/ROM behavior, watchdog/reset flags, enumeration, write/readback/return, and recovery are unverified on the target board.
- USB cables/topology, udev permissions, and appliance `dfu-util` behavior need target-host proof.
- Backend restart loses in-memory operation history; uncertain outcomes intentionally require reconciliation/ST-Link rather than optimistic cleanup.
- There is no automatic retry in this first implementation.
- Custom/expert builds remain ST-Link-only.

## Rollback / recovery

ST-Link/OpenOCD support is preserved. Keep ST-Link attached during HIL. If USB entry or application recovery fails, stop USB attempts, preserve logs, restore the backed-up/known-good image with ST-Link, verify flash byte-for-byte, reboot into the application, and only then deliberately release/restart update maintenance.


liltaket:
## First RM1000 hardware validation — blocked at ROM USB enumeration

I ran the first real bench validation on a Biltema RM1000 / RM-MB V6.1 (`Yardforce500B`, STM32F401VC) with the blade removed, wheels constrained, ST-Link attached, and no motion commands.

### What passed

- Two independent 256 KiB flash backups matched exactly (`fd012d05639f85944148142b6221b87f70c801260b385aae57c51ab5637dab20`).
- The ARM64 ROS and GUI HIL images ran on the Pi.
- OpenOCD with `reset_config none` reliably programmed and verified the F401 without NRST. This avoids the reproducible `st-flash` partial-write failure (`FLASH->SR 0x000000c0`, first failure at byte 11,892).
- Protocol-7 HIL firmware booted and completed exact runtime handshakes.
- The updater correctly enforced readiness, entered `waiting_dfu`, refused ambiguous identity, and never reached erase/write.
- ST-Link recovery remained reliable; HIL-A v1.10.234 was restored and OpenOCD reported `Verified OK` after the terminal test.

### What failed and what we learned

Initial direct-ROM experiments were reset by the application's 40 ms window watchdog during a blocking 500 ms USB-detach delay. The saved reset cause was consistently `RESET_CAUSE_WWDG=6`.

A final diagnostic image moved WWDG reset/gating before that delay. This removed the watchdog failure and proved that the CPU successfully transferred into STM32 system-memory ROM:

```text
PC   0x1fff0d4e
SP   0x20002140
xPSR 0x21000000
saved reset cause 0
```

However, Linux observed no new USB disconnect and no `0483:df11` device within the 20-second identity window. The old application identity `0483:5740` remained visible. The operation therefore terminated safely as `failed_requires_stlink / dfu_identity_uncertain`; no USB erase or write occurred.

This narrows the hardware blocker to the application-to-ROM USB handoff on RM-MB V6.1: host-visible detach, USB/RCC peripheral cleanup, or ROM interface selection. It is no longer an artifact-validation, updater-command, flash-tool, or ROM-vector problem.

The run also exposed a separate integration issue: degraded localization prevented `COMMAND_STOP=8` from reaching `StopHoldSequence`, so the controller could not publish IDLE for maintenance. A local fix exempts STOP from the localization guard while retaining blade-off and zero-velocity behavior. That fix should be reviewed independently before inclusion.

### Status

USB-DFU is **not hardware-functional on this RM-MB V6.1 setup yet**, and none of the required repeated update, interruption, physical-inhibit, or fault-injection campaigns can be claimed.

One bounded follow-up diagnostic is still worthwhile: enter the proven ROM state, then physically reconnect only the STM32 USB cable while watching for `0483:df11` on the same Pi port.

- If `df11` appears, time-box further work to a proper OTG soft-disconnect/full USB deinit and preferably the reset-marker/system-reset architecture.
- If it does not appear, stop the factory-ROM USB path for this hardware and keep OpenOCD/ST-Link as the supported update/recovery method until a separate bootloader design is undertaken.

No further firmware variants or cable cycles were attempted after this stop condition.

liltaket:
Parking this for now. The hardware validation was useful and narrowed the blocker down significantly, but getting application-to-ROM USB DFU reliable on this F401/RM1000 path turned out to be more involved than I had hoped. Closing the draft to keep the active PR list clean while leaving the investigation and results here for history/reference. I may come back to this later, possibly with a smaller or different bootloader/DFU approach.

Issue #637, opened 2026-09-16
mowgli_map: recorded areas have no stable ID — only fragile positional index


fwelvering:
### Component

ROS2 Stack

### Problem or Motivation

Recorded mow areas have **no stable identity** today — only positional index into `map_server_node`'s internal `std::vector<AreaEntry> areas_` (`ros2/src/mowgli_map/include/mowgli_map/map_server_node.hpp:355-361`). `MapArea.msg` has no `id` field, and every consumer that targets a specific area (`StartInArea.srv`, `GetMowingArea.srv`, the GUI's cross-hatch-per-area setting, coverage statistics) addresses it by that raw 0-based index.

This is fragile in a way that's worse than "deleting an area shifts later indices": the GUI implements editing or deleting a *single* area by tearing down and re-adding the **entire** list (`clear_map` → `add_area` × N in frontend order, see `gui/pkg/api/mowglinext.go` `replaceMapInternal` / `gui/pkg/msgs/mowgli/services.go`'s `ReplaceMapReq` doc comment) — so any single-area edit reshuffles every area's index, not just the ones after it.

This matters now because of planned external integrations (e.g. an MQTT/Home Assistant bridge, mowglinext-ha) that want to say "mow this specific area" reliably from outside the GUI session. Binding that to a positional index that can silently repoint at a different polygon after an unrelated edit is unacceptable — an external automation could end up mowing the wrong area with no indication anything changed.

There's also a currently-hidden internal risk from the same root cause: `mowgli_behavior`'s coverage-resume persistence (`ros2/src/mowgli_behavior/src/coverage_persistence.cpp`) — which area is `current_area`, per-area cross-hatch progress, per-area swath-completion bitset, the `~/start_in_area` single-area target — is **entirely index-keyed**, with no id fallback. If an area is edited (index reshuffle) while a mow is interrupted, resuming after restart can silently misattribute saved progress to the wrong area.

### Proposed Solution

Add a persistent, stable `id` to each recorded area, in the same additive style already used twice in this exact file for other per-entity fields:

- Per-obstacle identity fields (issue #502, `area_manager.cpp` obstacle `_name`/`_source` keys) — optional-on-read, only emitted when non-default.
- The datum stamp migration (issue #216, `migrate_areas_datum`) — an old `areas.dat` without the field gets it assigned and the file is re-saved on first load with the new binary.

Concretely:

1. **`MapArea.msg`**: add `uint32 id`.
[...]

PR #643, opened 2026-09-17
fix(mowgli_localization,mowgli_map): capture dock position before reversing, not after (#446)

### LINKED ISSUE #446: Docking calibration fails

fwelvering:
## Summary

Fixes #446 (the follow-up reported against it — full diagnosis in [this comment](https://github.com/mowglinext/mowglinext/issues/446#issuecomment-5710363301)). The one-click dock calibration's persist call has been failing 100% of the time since #621, with logs like:

```
07:13:11 WARN map_server_node: set_docking_point rejected: robot not detected on dock (is_charging=false, status_age=0.1s, max=3.0s). Drive onto the dock and wait for the firmware to report charging before retrying.
```
(repeating every ~5s)

#621 (correctly) moved the persist to run before the live re-dock confirmation — the re-dock steers on the OLD heading regardless of when the write happens, so deferring it just discarded good measurements whenever the old heading was too far off for the live approach to land. But that reorder exposed two problems neither #621 nor the original #446 accounted for:

1. **`persist_dock_via_map_server()` now structurally can never pass `on_set_docking_point`'s on-dock gate.** It fires right after the reverse leg's COG-coherence check — i.e. while the robot has deliberately driven off the dock. `on_set_docking_point`'s `is_charging` gate is unconditional regardless of `yaw_source`, so a `yaw_source=MOTION` call (measured off-dock by design, per `SetDockingPoint.srv`'s own doc comment) can never pass it.
2. **Even with that gate relaxed, the position would come out wrong.** The call sent `use_gps_position=true`, which tells `map_server_node` to average its own `recent_gps_antenna_enu_` — a window that keeps accumulating from `/gps/fix` with no on-dock gating at all. By persist time (after the reverse leg) that average is contaminated with in-motion samples from the very leg that just drove the robot away from the dock.

## Fix

- `calibrate_imu_yaw_node` now averages its own short window of raw, RTK-Fixed `/gps/fix` samples (mirroring `map_server_node`'s own antenna-averaging technique exactly, for the same anti-circularity reason issue #446/PR #610 already established: correcting with a possibly-stale fused yaw would bias a lever-arm-corrected average) **before the reverse leg starts**, while still confirmed on the dock.
- If `is_charging` drops or too few RTK-Fixed samples arrive within the averaging window, **the whole calibration run aborts right there** — reversing and measuring yaw would be pointless if the position capture that has to happen anyway already failed.
- `persist_dock_via_map_server()` sends that frozen, pre-reverse position explicitly (`use_gps_position=false`) instead of asking map_server to average its own (by-then contaminated) live window.
- `on_set_docking_point`'s `is_charging` gate is now skipped for exactly one combination — `yaw_source=MOTION` with an explicit position — since that position was already captured under verified on-dock conditions earlier in the same run. Every other combination (the live on-dock GPS-averaging path used by the GUI's simple "capture from robot" flow, `PRESERVE`, manual `REQUEST`) keeps the gate exactly as before.

## No interface/codegen changes

Everything above uses fields `SetDockingPoint.srv` already has (`yaw_source`, `use_gps_position`, `docking_pose.position`) — no `.msg`/`.srv`/`.action` field changes, so no codegen regeneration needed (same "No .action fields changed" discipline #621 already applied). The `CalibrateDock.action` header comment and both affected codemaps (`mowgli_localization.md`, `mowgli_map.md`) are updated to match the new flow.

## Field tested on physical hardware (2026-09-18)

Ran the full flow — on-dock position averaging → reverse + yaw measurement → persist → live re-dock confirmation — on a real robot (built from this branch merged into a fork's `dev`, CI-green):

- Position capture and persist succeeded on the first pass, no aborts on the `is_charging`/RTK-Fixed gate.
- The re-dock confirmation's first attempt landed ~0.4 m off cross-track at `opennav_docking`'s "contact claimed" (short of actual charging contact) — ordinary approach variance, not a regression from this change. `docking_server`'s own 10 s charge-timeout + 2 internal sub-retries handled it exactly as designed, then aborted the `DockRobot` goal after ~92 s.
- That correctly triggered this PR's re-dock supervisor's steered ~2 m backoff, re-lining the robot on the ideal dock line.
- The second attempt landed cleanly and started charging within ~23 s, no further retries needed — confirmed via `DockCalibrationStatus` and the `docking_server`/`behavior_tree_node` container logs, full sequence traced end-to-end.

This exercises exactly the failure path #446 originally reported (position must be captured on-dock, before the reverse) plus the re-dock confirmation this PR adds — both now verified working on hardware, not just by inspection.

## Test plan

- [x] `clang-format` (cached 18.1.8 binary + `git_clang_format.py`) — clean diff against `upstream/dev` for both changed C++ files.
- [x] Manual brace/paren balance check on both changed C++ files.
- [x] Python syntax check (`py_compile`) on the launch-file change.
- [x] YAML parse check on the template change.
- [x] Careful manual trace of the new pre-reverse capture path against the existing subscription lifecycle (`activate_sensor_subs`/`deactivate_sensor_subs`), confirming the new `/gps/fix` subscription and its accumulation window are cleared per-run so a stale window from a previous calibration attempt can't leak into the next one.
- [x] Field-tested on physical hardware — see "Field tested on physical hardware" section above; full calibrate → persist → re-dock-confirm flow validated end-to-end, including the steered-backoff retry path.
- [x] Build-verified via CI (this branch built and passed `Build & Test (ROS2 kilted)` after merging into a fork's `dev`) — not build-verified in the authoring sandbox itself (no local ROS2/colcon toolchain there). This PR still touches dock-pose persistence (root `CLAUDE.md` Invariant 6, safety-critical) — please review the gate-relaxation logic in `area_manager.cpp` `on_set_docking_point` particularly carefully regardless.


fwelvering:
Updated PR with fieldtest on physical mower

cedbossneo:
Thanks for this — the problem you identified is real and your diagnosis of it matches ours exactly: the dock position has to be captured while the robot is still ON the dock and charging, not after the reverse leg.

We hit the same failure on the robot on 2026-09-17 and had a fix in flight in #646, which we have now landed. Since both PRs rewrite the same six-step routine in the same files (`calibrate_imu_yaw_node.cpp`, `area_manager.cpp`, `CalibrateDock.action`), they cannot both apply, so I am closing this one in favour of #646. That is purely about which patch landed first, not about the quality of the analysis here.

Two differences are worth passing on, because the second one cost us a day of field debugging:

1. **Lever arm.** In this PR the averaged raw `/gps/fix` position is sent as `docking_pose.position` with `use_gps_position = false`, which is the "use the value as given" path. The `base = antenna − R(yaw)·lever_arm` correction in `area_manager.cpp` only runs inside the `use_gps_position` branch, so the persisted dock pose ends up being the ANTENNA position rather than `base_footprint` — off by the full lever arm (`gps_x`, 0.30 m on our robot). We chased exactly this: a stored pose 0.35 m short along the heading made `docking_server` claim contact at the right coordinates while the robot sat 0.35 m before the charge contacts, and `fusion_graph` logged a 0.35 m gauge reset on every successful charge. #646 keeps the correction in map_server (`dock_antenna_capture.hpp::DockBaseFromAntenna`, applied exactly once) and has a test for it. It is possible this also explains the ~0.4 m miss you reported in your own field test, though that was described as cross-track while the lever arm is mostly along-track, so I would not claim it without a bench measurement.

2. **Charging gate.** This PR relaxes the `is_charging` gate for `yaw_source == MOTION && !use_gps_position`, i.e. whenever a caller supplies a position. #646 keeps that combination gated and instead has map_server itself take the measurement, through a `~/capture_dock_antenna` service that runs the charging, RTK-accuracy and minimum-sample gates and holds a single-use, TTL'd pending capture. The intent is that no path can ever write a dock position that was not measured on the dock.

If you would like the phase-2/phase-3 work that rides on top of this branch in #658 reviewed on its own, splitting it out would make that much easier — right now #645 ⊂ #643 ⊂ #658 and the #637 phases are not mentioned in the #658 description.

cedbossneo:
Closing in favour of #646, which landed the same fix. Details in the comment above.

Issue #657, opened 2026-09-18
Obstacle drawn inside an area gets attached to a DIFFERENT area after splitting the area in two — robot mows straight through it


fwelvering:
## Summary

Field report: obstacles drawn inside a mowing area were mowed straight through by the robot. Investigation traced it to the obstacle being **persisted under the wrong area** — not a coverage-planning bug.

## Evidence

The operator had earlier split one mowing area into two ("Voor" and "Zij-Voor") using the map editor's split tool. Three obstacles, visually drawn well inside "Zij-Voor"'s boundary, turned out to be recorded under "Voor" instead:

```
area_2_name: Zij-Voor
area_2_obstacle_count: 0
area_3_name: Voor
area_3_obstacle_count: 3
```

(`docker exec mowgli-ros2 grep -E "^area_[0-9]+_(name|obstacle_count):" /ros2_ws/maps/areas.dat`)

`PlanCoverageArea`'s own dispatch log confirmed the planner never saw them:

```
PlanCoverageArea: dispatching — outer 46 pts, bbox=9.60x14.10m, signed_area=81.03m² (CCW=Y), perimeter=43.20m, 0 holes
```

## Root cause

`performSplit` (`gui/web/src/pages/map/hooks/useMapEditing.ts`) rebuilds the two resulting half-polygons when a workarea is split in two, but never re-evaluates which existing obstacles (previously children of the single pre-split area) now geometrically belong to which half. They all silently stay attached to whichever half kept the *original* JS object reference (`ObstacleFeature.mowing_area`), regardless of where the obstacle's own coordinates end up. An obstacle that geometrically ends up in the **new** half still renders correctly on screen — obstacles draw at their own coordinates, independent of which area's list references them — so there is **no visual indication** anything is wrong; only the persisted `areas.dat`/coverage plan is affected.

Three other obstacle-parenting call sites in the same file share a related but distinct bug: they pick a parent area via `Object.values(...).find(...)`, i.e. the **first** candidate. Two of the four (splitting an obstacle polygon itself, and changing an existing feature's type to "obstacle") do this with **no geometry check at all**; the other two (`addObstacle`, `onCombine`'s obstacle case) do check containment but still take the first match when more than one area's polygon legitimately contains the point — which is exactly what two adjacent/touching post-split areas produce near their shared edge.

## Fix

[...]

PR #658, opened 2026-09-18
fix(gui): re-parent obstacles by geometry, not first-match, at 4 map-editor sites

fwelvering:
Fixes #657.

## The bug

An obstacle drawn inside a mowing area can end up persisted under a **different** area — with no visual indication anything is wrong, since obstacles render at their own coordinates regardless of which area's list references them. The coverage planner then never sees it as a hole (`PlanCoverageArea` dispatch log showed `0 holes` for the affected area every time), so the robot mows straight through it.

Confirmed on a real map: `areas.dat` showed 3 obstacles drawn inside "Zij-Voor" recorded under "Voor" instead (`area_2_obstacle_count: 0` / `area_3_obstacle_count: 3`). "Voor"/"Zij-Voor" were produced by splitting one original area into two.

## Root cause

`performSplit` (`gui/web/src/pages/map/hooks/useMapEditing.ts`) rebuilds the two resulting half-polygons when a workarea is split, but never re-evaluates which existing obstacles (previously children of the single pre-split area) now geometrically belong to which half — they all silently stay attached to whichever half kept the *original* JS object reference (`ObstacleFeature.mowing_area`).

Three other obstacle-parenting sites share a related bug: they pick a parent via `Object.values(...).find(...)` — the **first** candidate. Two do this with **no geometry check at all** (splitting an obstacle itself; changing a feature's type to "obstacle"); the other two (`addObstacle`, `onCombine`'s obstacle case) check containment but still take the first match when more than one area's polygon contains the point — exactly what two adjacent post-split areas produce near their shared edge.

## Fix

- New shared `findContainingArea()` helper: checks every candidate area and prefers the **smallest** (most specific) containing one instead of the first found. Two areas that touch/overlap at a shared edge can both legitimately contain a point; silently picking "whichever came first" is exactly the bug.
- Used consistently at all 4 obstacle-parenting sites (`addObstacle`, `onCombine`, `performSplit`'s own obstacle-split case, `updateMowingArea`'s type-change case).
- `performSplit`'s **workarea** case gets the missing behavior entirely: after building the two new half-polygons, every obstacle that belonged to the pre-split area is re-evaluated against both halves and re-parented to wherever it actually ends up. An obstacle matching neither half (e.g. the cut line ran through it) stays on the original half rather than being guessed at — that's the pre-split status quo, not a new misattribution.

## Field tested on physical hardware (2026-09-18)

Reproduced and fixed the exact reported scenario end-to-end on the reporter's own robot:

- Restored a backup of `areas.dat` from before the original area was split — at that point the 3 obstacles were confirmed correctly attached to the single, not-yet-split area (`obstacle_count: 3`).
- Updated `mowglinext-gui` to this fix and repeated the same split operation that originally caused the bug.
- Re-checked `areas.dat`: the 3 obstacles landed on the correct resulting half (the one they are geometrically drawn inside — `Voor-Zij`, `obstacle_count: 3`) instead of the unrelated other half (`Voor`, `obstacle_count: 0`), confirming `performSplit`'s new re-parenting pass works correctly on a real, previously-broken map, not just the synthetic unit-test geometry.

## Test plan

- [x] New unit tests: `findContainingArea`'s smallest-match / order-independence behavior (3 cases).
- [x] New end-to-end test driving the real `performSplit` path (via `handleSplit` + `onCreate`) on a 10x10 square cut down the middle, confirming an obstacle that geometrically ends up in the new half gets re-parented there, and a control confirming one that stays in the original half is left alone.
- [x] Manual brace/type review of all 4 changed call sites — `findContainingArea`'s candidate/return types match every call site's existing usage (`MowingAreaFeature[]` in, `MowingAreaFeature | undefined` out).
- [x] Field-tested on physical hardware — see "Field tested on physical hardware" section above; reproduced the original bug from a pre-split backup and confirmed the fix re-parents obstacles correctly on a real, previously-affected map.
- [x] CI green (`gui-ci.yml`: tsc + eslint + vitest) on the fork after merging — not run locally in the authoring sandbox (no Node/yarn toolchain there).


fwelvering:
Field-tested today on our fork (fwelvering/mowglinext, synced through `fork/dev`). Confirmed working as designed:
- Split a workarea with an obstacle inside it — the obstacle was correctly re-parented to the geometrically containing half, not left on the original.
- Drew a new obstacle near a shared boundary between two adjacent areas — assigned to the smallest/most specific containing area, not the first match.
- Changed an existing feature's type to obstacle inside a specific area — correctly parented.

No misattribution observed at any of the tested sites.

fwelvering:
Rebased against the latest `dev` and resolved the merge conflicts (`bt_context.hpp`, `behavior_tree_node.cpp`, `coverage_nodes.cpp` — all additive, two independent features landing new fields/subscriptions side by side; also corrected a stale comment in `coverage_nodes.cpp` that no longer matched actual `attempted_areas`/`completed_areas` behavior). CI is running on the updated commit; all checks green so far except ROS2 — CI / ROS2 — Docker, still in progress.

Issue #659, opened 2026-09-18
Pause then Resume on a targeted single-area run restarts at area 0 instead of continuing the paused area


fwelvering:
## Summary

A targeted single-area run (started via the GUI’s “mow this area” button, e.g. on area 3 “Voor”) that is interrupted with **Pause** (`COMMAND_STOP`, `StopHoldSequence`, `state_name="IDLE"`) restarts at **area 0** when the operator presses **Resume**, instead of continuing the paused area.

## Root cause

`clearSingleAreaMode()` is called unconditionally on every plain `COMMAND_START` received by the `~/high_level_control` handler (`behavior_tree_node.cpp`). This exists to stop a stale single-area clip from surviving a low-battery-dock/emergency auto-resume, where the mowing session legitimately stays alive without `EndSession` running — in that case the operator’s next Start explicitly does expect the whole lawn, so the unconditional clear is correct there.

But the same unconditional clear also fires on the far more common case: a manual Pause mid-run, immediately followed by Resume within the same interrupted session. There the operator expects Resume to continue exactly what was paused, single-area target included — not silently widen to the whole lawn.

## Reproduction

1. Start a targeted single-area run via the GUI’s “mow this area” button on any area.
2. Press **Pause** partway through.
3. Press **Resume** (or “Start” — same `COMMAND_START` under the hood, see `MowerActions.tsx`).
4. Observe: `GetNextUnmowedArea` logs `"area 0 selected"` instead of re-entering `"targeted run — mowing only area N (single-area mode)"`.

Field log excerpt:
```
GetNextUnmowedArea: targeted run — mowing only area 3 (single-area mode)
GetNextUnmowedArea: area 3 selected (0 swath(s) done so far) — dispatch attempt 1/5
... [Pause, then Resume] ...
GetNextUnmowedArea: area 0 selected (0 swath(s) done so far) — dispatch attempt 1/5
```

## Fix

See linked PR: add `isResumableHoldState()`, true only for plain `"IDLE"` (the state `StopHoldSequence` publishes, and the only branch that does so without also being reachable from a charge hold or a session end), and skip `clearSingleAreaMode()` when it holds. The charge-hold/emergency scenario the original clear was written for is unaffected — that publishes `CHARGING`/`CRITICAL_BATTERY_CHARGING`, never plain `IDLE`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

fwelvering:
Fixed in @660

[...]

PR #660, opened 2026-09-18
fix(mowgli_behavior): preserve single-area target across a plain Pause/Resume

fwelvering:
Fixes #659.

## The bug

A targeted single-area run (started via the GUI’s “mow this area” button) that was interrupted with **Pause** (`COMMAND_STOP`, `StopHoldSequence`, `state_name="IDLE"`) restarted at area 0 on **Resume**, instead of continuing the paused area.

## Root cause

`clearSingleAreaMode()` was called unconditionally on every plain `COMMAND_START` received by the `~/high_level_control` handler. That is correct for the low-battery-dock/emergency auto-resume case it was written for (the session stays alive without `EndSession`, and the operator’s next Start does expect the whole lawn there) — but the same unconditional clear also fired on the far more common manual Pause → Resume within the same interrupted session, where Resume should continue exactly what was paused.

## Fix

Adds `isResumableHoldState()` (`bt_context.hpp`), true only for plain `"IDLE"` — the state `StopHoldSequence` is the sole publisher of outside a charge hold or a session end — and skips `clearSingleAreaMode()` when it holds. The original charge-hold/emergency scenario is untouched: it publishes `CHARGING`/`CRITICAL_BATTERY_CHARGING`, never plain `IDLE`.

## Trade-off

This removes the implicit “escape hatch” Resume used to offer for switching a paused targeted run back to full-lawn mode. To explicitly mow the whole lawn after pausing a targeted run, the operator now needs to start a different area’s “mow this area” button (which always sets a fresh target) rather than relying on Resume’s old side effect. Flagging this for reviewer awareness; happy to add an explicit “mow whole lawn instead” affordance in a follow-up if wanted.

## Test plan

- [x] New unit test `ResumableHoldStateIsExactlyPlainIdle` (`test_manual_resume.cpp`)
- [ ] Field-tested on physical hardware — pending a `mowgli-ros2` rebuild from this branch; will update this PR once confirmed
- [ ] Not built/run locally in the authoring sandbox (no C++ toolchain there) — relying on CI plus the field test above


cedbossneo:
Thanks for this. The bug is real and the diagnosis is clean: `clearSingleAreaMode()` firing on every plain `COMMAND_START` is wrong for a manual Pause → Resume inside the same session, and gating it on the `"IDLE"` hold state is a sensible, narrow fix.

Before it can land, though, this branch needs to be rebased on current `dev`, for two reasons.

**1. It carries #643, which we have just closed.** `git merge-base --is-ancestor pr-643 pr-660` is true: this branch includes `007df142` "capture dock position before reversing, not after" along with the rest of the stack (#637 phases 1–3 and the #658 re-parenting). #643 was closed in favour of #646 — see the comment there for the detail, but the short version is that it sends the averaged raw `/gps/fix` position through the `use_gps_position = false` path, where the lever-arm correction never runs, so the persisted dock pose is the ANTENNA position rather than `base_footprint`. We spent a day on the robot chasing exactly that symptom (dock pose 0.35 m short, `docking_server` claiming contact 0.35 m before the charge contacts). Merging this PR as it stands would put that back.

**2. It now overlaps heavily with #626, which merged today.** `coverage_nodes.cpp`, `bt_context.hpp`, `coverage_persistence.*` and `test_get_next_unmowed_area.cpp` all changed substantially: FollowStrip gained same-unit resume, dig skip zones and action-result handling, and the path-divergence guard was removed. The conflicts will be real rather than textual, so they are better resolved by you, with the intent of this fix in mind, than by us guessing.

Two suggestions that would make this much faster to review and land:

- Rebase onto `dev` and keep **only the Pause/Resume fix** in this PR. At 1378 added lines across 43 files, most of what is here is the stack underneath rather than the fix described in the title. The fix itself looks small.
- #645 (stable area ids, phase 1) is about to land on its own — I pushed a small follow-up to it that refuses a round-tripped id another area already holds, since phases 2 and 3 key on that id. Once it is in, phases 2 and 3 can be rebased on top as their own PRs.

Side note, since your branch tip is a CI re-run for it: `GetNextUnmowedAreaTest.OrientationServiceEditsNextWithoutChangingActivePlan` is flaky on the runner for us too. It failed once on #626 and passed on a plain re-run with no code change. It depends on a 10 ms wall timer ticking `processPending()` during a service call, which is the sort of thing that slips on a loaded runner. Not yours to fix here, but worth knowing it is not your change.

fwelvering:
Thanks for the detailed review — rebased as suggested.

- Re-created this branch from current `dev` (post-#626), then cherry-picked only the original fix commit (`fc789bfe`) on top — no more #643/#645-#637-phase-2-3/#658 stack riding along.
- The cherry-pick applied cleanly with no conflicts: the addition (`isResumableHoldState()` plus the guard at the `~/high_level_control` call site) lives in a part of `bt_context.hpp` / `behavior_tree_node.cpp` that #626 didn't touch, so there was nothing to reconcile against the same-unit-resume/dig-skip-zone changes there.
- Diff is now 3 files / +60 −1, matching the original fix exactly (`isResumableHoldState()` + the guarded `clearSingleAreaMode()` call + the new unit test).
- Force-pushed the new head; CI is re-running on it now.
- Noted re #643/#646 and the flaky `OrientationServiceEditsNextWithoutChangingActivePlan` test — appreciate the context, useful to know both aren't things to chase here.

Will follow up once this is field-tested against the reported Pause/Resume scenario.

PR #674, opened 2026-09-19
feat(gui): make map area selection and persistence area-id aware (#637, phase 3/3)

fwelvering:
Phase 3 of 3 for #637 (stable per-area id), rebased on current `dev` (post-#645/phase 1, post-#626) and scoped to only the GUI-side consumer -- no dock-calibration or other stack content.

## The problem

A mowing area's ROS array index (`working_area_indices`) is NOT a stable identity: map_server rebuilds the whole area list (clear + re-add) on any edit save, so a position captured at render time can silently point at a DIFFERENT area by the time an action fires against a later `map` poll -- exactly the live, mid-session case the GUI's own edit/delete flow produces.

## Fix

- New `mowingAreaIndexById` (`utils/mapAreaIndex.ts`): resolves the CURRENT ROS index from an area's stable id (`MapArea.id`, phase 1) instead of a captured position.
- `MapPage.tsx`: `obstacleAreaIndex`/`obstacleAreaNames` and `startSelectedArea` now resolve by id at the moment they fire, not from a position captured whenever `features`/`mowingAreas` last rebuilt (frozen while `editMap` is true). `buildFeatures` stamps `source_working_area_id` alongside the existing `source_working_area_index`; `useMapEditing.ts`'s `buildLabels` carries it onto the centroid label feature `mowingAreas` is built from.
- `useMapFiles.ts`: `handleSaveMap` round-trips each area's existing id in the save payload so map_server's `on_add_area` can preserve it across the wholesale rebuild, instead of silently minting a fresh one on every save -- which would discard BT coverage-resume progress for every area, not just the one being edited.
- Backend: `MowgliMapArea` gains `id` (hand-added to the swagger docs + `Api.ts` since nothing in the repo runs `swag`; `splitMapAreas`/`pollMap` already pass every `MapArea` field through untouched, so no other backend wiring was needed).

## Test plan

- [x] New/extended unit tests: `mowingAreaIndexById` (order-independence, deleted-id, non-integer id, an area that moved position), `useMapFiles` save-payload id round-trip (existing id sent back, omitted for a brand-new area), `splitMapAreas` preserves the id alongside the ROS index.
- [ ] Not built/run locally in the authoring sandbox (no Node/yarn/Go toolchain there) -- relying on CI.
- [ ] Field-tested on physical hardware -- will update this PR once confirmed against a live area edit/save cycle.


fwelvering:
Field-tested today alongside #658. Opened a specific area's edit panel in the map editor, then performed an edit elsewhere that changed the area list (add/reorder) while that panel stayed open. It correctly kept referencing the same area (name/polygon/obstacles unchanged) instead of snapping to whatever ended up at the old list position. Confirmed working as designed.

PR #681, opened 2026-09-19
feat(mowgli_behavior,mowgli_map): make coverage-resume state area-id aware (#637, phase 2/3)

fwelvering:
Phase 2 of 3 for #637 (stable per-area id), rebuilt from scratch on current `dev` (post-#645/phase 1, post-#626) after the original branch (based on stale `fork/dev`) was flagged as unmergeable in #660's review.

## The problem

The BT's coverage-resume tracking (`GetNextUnmowedArea`, `coverage_persistence`) keys every per-area map (`completed_areas`, `attempted_areas`, `area_completed_swaths`, `area_resume_pose_index`, `cross_hatch`, ...) by the area's ARRAY INDEX, not the stable id from phase 1. The GUI's area edit/delete flow rebuilds the whole area list (`map_server`'s `on_add_area`: `clear_map` + one `add_area` per surviving area), which can shift what area a given index refers to, live and mid-session or across a restart. An index that inherited another area's stale "completed" flag would be skipped forever while genuinely unmowed, or hand a re-indexed area another area's swath/cross-hatch history.

`GetNextUnmowedArea` now reconciles every probed index's `MapArea.id` against the last one observed for that index (`ctx->area_ids`, persisted in `coverage_resume.txt` as format v3) on every probe. A mismatch discards the per-index state for that slot.

## Design conflict found against #626, and how it's resolved here

#626 added a synchronous fast-skip path to `onStart()`/`advanceAndProbe()` (skip already-completed/attempted indices without firing a probe, for performance). That directly undermines the id-reconciliation guarantee: an index skipped this way is never re-verified, so an edit that lands a different, genuinely unmowed area on that index between two `GetNextUnmowedArea` entries (e.g. Pause, edit the map, Resume — no probe ever fires in between) would silently skip it for the rest of the session. The original phase-2 branch (authored before #626 existed) resolved this by removing the fast-skip path outright.

This version keeps #626's fast-skip path but gates it: `map_server` now bumps and publishes a generation counter (`~/area_list_generation`, `std_msgs/UInt64`, transient_local) on every successful `~/add_area`, subscribed by `behavior_tree_node` into `BTContext::current_area_list_generation` (topic-driven, so it is current regardless of `GetNextUnmowedArea`'s own probe cadence). Each probed index also records the generation it was verified at (`area_verified_generation`). The fast-skip is only taken when an index's verified generation matches the current one — i.e. nothing has changed since it was last actually probed. An unverified index falls through to a real probe, which re-verifies it via the existing id-reconciliation and updates both maps. This keeps #626's optimisation for the common case (nothing changed) while restoring phase 2's original safety guarantee for the case that actually matters.

## A second, field-found gap: `single_area_target` itself

Field-testing this on the fork surfaced a related bug the id-reconciliation above does NOT cover: `single_area_target` (the `~/start_in_area` "mow only THIS area" clip) is stored as a raw INDEX. Targeting area 0 (id 6), pausing, reordering the area list so a different area (id 7) ended up at index 0, then Resume — the reconciliation correctly discarded id 6's stale per-index state for index 0, and then dispatched index 0 anyway, since `single_area_target` had nothing to check the id against. The robot silently started mowing an area the operator never selected.

Fixed by locking `single_area_target_id` in from the first successful probe after the target is set, verified on every later probe of that same index; a mismatch ends the run (`FAILURE`, `coverage_all_complete` left false so it routes to the failure/dock path) rather than silently mowing the wrong area. Deliberately not a scan-and-relocate across the whole area list — failing safe and asking the operator to re-select the area is a small UX cost next to the risk of running blades in an unintended zone. No GUI/interface/codegen change needed: the id comes from the existing probe response, not a new field threaded through `~/start_in_area`.

## Test plan

- [x] Unit tests for the id-reconciliation (`IdMismatchClears...`, `MatchingIdKeeps...`, `NoRecordedIdMeans...`), the fast-skip generation gating (`FastSkipNeverTrustsAnIndexNeverProbedThisProcess`), the `single_area_target_id` anchor (`TargetedRunLocksInAreaIdOnFirstDispatch`, `TargetedRunEndsCleanlyWhenTargetMovedToADifferentIndex`, `TargetedRunContinuesNormallyWhenTargetIdStillMatches`, `FreshTargetRequestDoesNotCompareAgainstAPriorTargetsId`), persistence round-trips (`area_ids`, `single_area_target_id`), and `map_server`'s `area_list_generation_` bump (`AddAreaBumpsTheGenerationCounter`).
- [x] Field-tested on physical hardware: reproduced the original id-reconciliation gap, the generation-gated fast-skip, AND the `single_area_target` bug live (Pause → reorder → Resume), confirmed each fix against the field logs before merging into the fork's own `dev`.
- [ ] Not built/run locally in the authoring sandbox (no C++ toolchain there) — relying on CI (green on the fork after 3 follow-up fixes for a formatting nit and two real bugs the CI/field-testing round caught) plus the field test above.


fwelvering:
**Field test results (physical mower):**
- Pause / resume: ✅ works
- Edit / reorder areas: ✅ works
- Pause → edit/reorder → resume (the exact scenario this PR's area-id reconciliation targets): ✅ works

fwelvering:
Rebased against the latest `dev` and resolved the merge conflicts (`docs/claude/codemaps/mowgli_map.md`, `bt_context.hpp`, `behavior_tree_node.cpp`, `coverage_nodes.cpp`). The docs conflict reflected #637 phases 1 and 3 having landed upstream in the meantime — updated the codemap to note all three phases are now in. The code conflicts were additive (new fields/subscriptions side by side); also corrected the same stale `attempted_areas` comment as in #658. CI is running on the updated commit; all checks green so far except ROS2 — CI / ROS2 — Docker, still in progress.

PR #684, opened 2026-09-19
fix(gnss): keep diagnostic storage updater-safe

wjcloudy:
## What

- replace the new `universal_gnss_logs` named volume with a bounded 64 MiB tmpfs
- retain bounded Docker `json-file` logs for normal diagnostics
- add an updater regression that permits the GNSS sidecar to join a pre-sidecar layout while continuing to reject unreviewed persistent mounts
- pin the storage contract in the installer shell test and document the persistence boundary

## Why

#679 made the Universal GNSS sidecar independent of installer-generated variables and files, but its new writable named volume still trips the layout-1 updater guard:

`gps adds writable storage at /var/log/universal_gnss; explicit data/layout migration required`

The exports are diagnostic scratch data, so this PR keeps them ephemeral. If persistent GNSS exports become a requirement later, that needs a separate layout migration plus backup/restore coverage.

## Upgrade behaviour and future benefit

This is not only a workaround for the first affected release. The failure occurs at the transition boundary: a pre-sidecar installation cannot accept the first stack that introduces a new writable named volume. A machine installed directly on a release where that volume already exists would generally regard it as existing storage on later updates, but that does not make the original transition safe or portable.

Using bounded tmpfs for disposable sidecar diagnostics makes the pre-sidecar to sidecar transition updater-safe now, and gives later releases a reusable pattern when adding similar scratch-only services. It does not weaken the layout guard: any future sidecar or stack change that introduces genuinely persistent storage must still provide an explicit migration, backup, and rollback contract before the updater will accept it.

Existing installations that already created `universal_gnss_logs` can update to this layout; the old Docker volume becomes unused and is deliberately not deleted automatically.

## Verification

- `go test ./pkg/updater`
- `bash install/tests/test_gnss_external_sidecar_contract.sh` — 21/21 passed
- `git diff --check`

### ARM64 hardware validation — 19 September 2026

Validated on Pi 5 hardware at PR commit `54172c25`, using the published ARM64 GNSS image pinned by the fragment (`v0.1.4-rc2`, digest `24ea6c1c0553463207a4c33b803e920a974989f5891f2f107e8c35cce56f7a95`, driver revision `c56356c2`).

- Production updater storage checks accepted the actual rendered tmpfs layout and downloaded image; the previous named volume was still rejected with the migration error.
- The non-root runtime could write logs/exports, the 64 MiB cap was enforced, and exports disappeared after container recreation. Docker log rotation remained configured at 3 × 5 MiB; no persistent volume was created.
- The actual sidecar launcher communicated with the installed receiver and reported healthy on initial start and after recreation: 76 and 75 distinct fix timestamps in two 15-second observations.
- Original services were restored and fresh GNSS/IMU telemetry verified, with the mower stationary and no motion commands issued.

These checks validate the storage change and stationary receiver startup. They do not claim a complete release installation/rollback or moving-mower field acceptance, which remains `HARDWARE_REQUIRED`. The exact test used 500B firmware `e6ced900` / 1.11.17 and the installed Unicore receiver; receiver hardware/firmware revision was not independently recorded. Raw evidence and baseline manifests are retained.


wjcloudy:
@cedbossneo could you please review this updater-safe storage change and merge it if it looks good? The PR description now clarifies both the affected upgrade boundary and the benefit for future scratch-only sidecars.

cedbossneo:
Closing as superseded by #687, which landed on dev before this one and fixes the same updater refusal ("adds writable storage … explicit data/layout migration required") by moving the sidecar's logs and exports into the existing `/run/universal_gnss` tmpfs rather than adding a second tmpfs at `/var/log/universal_gnss`. The two approaches conflict in `docker-compose.gps.yml` and the contract test, and dev already pins the property in `gui/pkg/updater/gps_fragment_test.go` and `test_gnss_external_sidecar_contract.sh`. Thanks for chasing it — the one thing here dev does not cover is the "gps joins a stack that had none" case in your Go test; happy to take that as a small standalone PR.

wjcloudy:
Thanks — I extracted the remaining regression case into #691, based on current `dev` and limited to the single Go test. It covers a pre-Universal-GNSS stack accepting the new ephemeral sidecar while retaining the migration guard for newly introduced persistent storage.

PR #688, opened 2026-09-19
feat(mqtt): add Home Assistant auto-discovery

wjcloudy:
The MQTT bridge exposes mower state and commands, but Home Assistant users otherwise have to define every entity by hand. This adds an opt-in **Home Assistant auto-discovery** setting under Settings → MQTT / Home Assistant. When enabled, the mower appears as one Home Assistant device with its native mower controls, selected-area actions, state, battery, GPS, charging, emergency, rain, and useful motor/power sensors.

## What it supports

- A native lawn-mower entity with start, pause, and dock controls.
- One explicit **Mow <area name>** button for each current mowable map area.
- Battery, coverage, GPS quality, RTK state and GPS location.
- Charging, emergency and rain indicators.
- Blade speed/current, battery voltage and charge current.

Home Assistant's built-in lawn-mower entity only defines start, pause, and dock. Area actions therefore appear as buttons under the same MowgliNext device, where they can also be placed beside the mower card on a dashboard. A dropdown was deliberately avoided because an MQTT select sends its command as soon as its value changes; an area choice should not unexpectedly start a physical mower.

## How it works

- Publishes one retained device-discovery document at `homeassistant/device/<derived-id>/config`.
- Reuses the existing `<prefix>/available`, telemetry, `<prefix>/command`, and `<prefix>/start_area` topics; no parallel control path is introduced.
- Polls the existing mowable-area service and refreshes discovery when areas are added, renamed, reordered, or removed. Area discovery IDs include both current index and name so a map reorder cannot silently retarget an existing Home Assistant automation.
- Republishes discovery after MQTT reconnect and when Home Assistant announces `online` on `homeassistant/status`.
- Services the MQTT network loop independently of the telemetry publish rate, keeping commands and discovery responsive even when telemetry is configured at a slow rate.
- Clears the retained discovery document for the current prefix when auto-discovery is disabled.
- Derives stable device/entity IDs from `mqtt_topic_prefix`, allowing multiple mowers on one broker when each has a unique prefix.
- Keeps discovery disabled by default and uses Home Assistant's standard `homeassistant` discovery prefix.

The setting is wired through the GUI schema, persisted mower config, launch injection, bridge defaults, English/French labels, and operator documentation.

## Validation

- `mowgli_monitoring`: 6 test targets passed, including 42 MQTT bridge tests; 0 failures.
- Launch injection tests: 15 passed.
- Template/runtime config drift check passed.
- JSON schemas and translation files parse successfully.
- Changed ROS 2 C++ files pass clang-format 18.
- Validated on Pi 5 hardware against an external MQTT broker: 17 discovered components, including four real map-area buttons. Clearing discovery and publishing the Home Assistant birth message restored all 17 components in three consecutive checks.
- GUI and ROS 2 containers remained running; GUI returned HTTP 200.

No mowing command was sent during discovery validation.


wjcloudy:
Selected-area mowing is now included in this PR.

- Each current mowable map area is discovered as an explicit `Mow <area name>` button under the same MowgliNext Home Assistant device.
- Buttons reuse the existing `<prefix>/start_area` command and current area index; navigation-only areas are excluded.
- Discovery refreshes when the area list changes. IDs include index and name so a reordered map cannot silently redirect an existing automation.
- Home Assistant's standard mower entity cannot contain custom area actions, so these are sibling entities on the same device. I avoided an MQTT select because changing a selection would immediately issue a command and start the mower.
- MQTT network servicing is now independent of the telemetry rate, fixing delayed discovery refresh after Home Assistant birth messages.

Validation on Pi 5 hardware with an external broker:

- 17 discovery components, including four live map areas (`Back`, `Front Side`, `Front Main`, `Test`).
- Three consecutive clear → Home Assistant `online` checks restored all 17 components and four area buttons.
- `mowgli_monitoring`: 6 test targets, 0 errors/failures; 42 MQTT bridge tests passed.
- GUI returned HTTP 200 and the mower containers remained running.

No area-start command was sent during this validation.

Issue #694, opened 2026-09-20
Universal GNSS sidecar (RC2): /gps/fix silently freezes, robot's actual localized position becomes wrong and stays wrong — not just a display issue


fwelvering:
## Summary

**This is not a map-display bug — the robot genuinely does not know where it physically is.** On a real robot running the Universal GNSS sidecar migration (#625/#679/#687, image `ghcr.io/pepeuch/universal-gnss-ros2-lyrical:v0.1.4-rc2@sha256:24ea6c1c0553463207a4c33b803e920a974989f5891f2f107e8c35cce56f7a95`), `/gps/fix` (`sensor_msgs/NavSatFix`, published directly by `receiver_node`, `-r fix:=/gps/fix`) repeatedly **freezes**: the message keeps publishing on schedule — `header.stamp` advances normally — but `latitude`/`longitude`/`status` stop changing entirely, for minutes at a time, while the robot is physically driving.

This reaches `fusion_graph` as an apparently-valid, apparently-fresh GNSS observation (fresh timestamp, well-formed message) that just happens to never move — which is a worse failure mode than an outright dropped/stale topic, because none of the existing staleness/age-based checks catch it: the topic **is** publishing, on time, every cycle. The fused pose — the SAME `map→odom→base_footprint` transform that navigation, boundary safety, and coverage all trust as ground truth, not just a display value — gets pinned near wherever the freeze started and never tracks the robot's real movement, even after `clear_graph` + driving several metres with confirmed-good wheel odometry flowing. The GUI map is only where a human *notices* this; the actual fault is that the localizer's internal belief about the robot's position has permanently diverged from reality, with the system reporting high confidence (tight covariance) in that wrong belief the whole time.

## Impact (updated — confirmed blocking, not intermittent)

**This currently makes mowing impossible, not just visually confusing.** Field-confirmed: the freeze recurs almost immediately after each recovery (within the time it takes to start a session), and because the coverage/navigation stack drives off the same wrong, falsely-confident fused pose, the robot does not follow its planned path — observed driving in a single straight line regardless of the actual planned coverage area. Combined with root CLAUDE.md's Invariant 16 finding that `fusion_graph`'s own marginal covariance is already known to be unreliable/overconfident, an operator has **no in-UI signal** that any of this is happening: GPS shows a fix, the graph reports tight covariance, and the robot drives with apparent purpose — just not where it's supposed to.

## Evidence

[...]

PR #701, opened 2026-09-21
fix(fusion_graph): stuck-gate rotation accumulator must reset per GPS message

fwelvering:
## Summary

Follow-up to #695. Field session `gnss-rc1-test-20260921-0933` (46.86 s, on `v0.7.0-rc1` via #700) surfaced a second design bug in the #694 stuck-value mitigation itself: `gps_rejects_stuck_value` latched 3 times right at the start of the drive and then **never again**, even though `/gps/fix` stayed frozen bit-exact for the entire session (confirmed in the same trace — see the [issue #694 comment](https://github.com/mowglinext/mowglinext/issues/694#issuecomment-5757245903)).

## Root cause

`abs_dtheta_since_gps_value_changed_rad_` (the gate's turn-exclusion stand-down accumulator) was reset with the *same* condition as the distance accumulator — only when the reported GPS value actually changes. Once `/gps/fix` froze, that condition never fired again for the rest of the drive.

A single transit turn starting at t≈24s integrated `|dtheta|` past 170°. Because the accumulator never got to reset, it permanently exceeded the 60° stand-down threshold and silently disabled the whole gate for the remainder of the stuck period — right through to the emergency stop.

## Fix

Renamed to `abs_dtheta_since_last_gps_sample_rad_` and reset it on **every** GPS message (accept or reject) instead — same philosophy `rtk_wrongfix_gate.hpp` already uses for its own accumulators. It answers "is a turn happening close to *this* sample", not "has there been a turn since the receiver got stuck" — the distance accumulator alone carries the latter, unbounded, semantics and is unchanged.

At this session's field-confirmed `/gps/fix` publish rate (~1 Hz — 46 distinct stamps over 44.95 s), a single inter-fix interval's rotation stays well under the stand-down budget even during a fast turn, so detection now continues through a turn instead of going silent for the rest of the stuck period.

## Scope

This only restores the #695 mitigation's own reliability during a turn. The underlying Universal GNSS receiver freeze (#694) remains unresolved and unrelated to this repo — tracked there.

## Test plan

- [x] New regression test `CumulativeRotationOverStuckPeriodDoesNotDisableDetection` reproduces the field-observed shape (cumulative rotation across a stuck period must not permanently disable detection) and fails without the fix
- [x] New test `ASingleViolentSingleMessageRotationStandsDownThatMessageOnly` pins the stand-down's actual intent (one implausible single-interval rotation stands down only that message, not the ones after it)
- [x] Existing tests updated to simulate the caller's real per-sample reset behaviour
- [ ] CI (`fusion_graph` unit tests + format)


Pepeuch:
Thanks for the field feedback — this is very useful.

The frozen /gps/fix while messages and timestamps are still advancing is especially interesting, because it means a simple “receiver/transport is alive” check is not enough.

I’m going to investigate this on the Universal GNSS side as well. Even if the root cause turns out not to be inside UG, this kind of stale navigation payload can be harmful because downstream consumers may continue to see apparently fresh messages while the actual position is no longer updating.

I’m considering adding native navigation-freshness diagnostics in UG, separating:

* transport/RX alive
* receiver solution time progressing
* position payload actually changing
* prolonged identical/stale navigation data

The goal would be to expose the condition cleanly without falsely declaring a stationary receiver faulty.

Your gnss-rc1-test-20260921-0933 session is a very useful real-world regression case for this. Thanks for testing v0.7.0-rc1 in the field.

fwelvering:
> Thanks for the field feedback — this is very useful.
> 
> The frozen /gps/fix while messages and timestamps are still advancing is especially interesting, because it means a simple “receiver/transport is alive” check is not enough.
> 
> I’m going to investigate this on the Universal GNSS side as well. Even if the root cause turns out not to be inside UG, this kind of stale navigation payload can be harmful because downstream consumers may continue to see apparently fresh messages while the actual position is no longer updating.
> 
> I’m considering adding native navigation-freshness diagnostics in UG, separating:
> 
>     * transport/RX alive
> 
>     * receiver solution time progressing
> 
>     * position payload actually changing
> 
>     * prolonged identical/stale navigation data
> 
> 
> The goal would be to expose the condition cleanly without falsely declaring a stationary receiver faulty.
> 
> Your gnss-rc1-test-20260921-0933 session is a very useful real-world regression case for this. Thanks for testing v0.7.0-rc1 in the field.

Happy to assist.
If it helps to share the complete session logs, let me know.

